PR #10, opened 2024-06-20
ci: retrieving snk for sdk/client docs signing

tanderson-ld:


.github/workflows/publish-docs.yml
```diff
@@ -15,14 +15,24 @@ jobs:
...
       - name: Setup Env from project's Env file
         shell: bash
         run: echo "$(cat ${{ inputs.pkg_workspace }}/github_actions.env)" >> $GITHUB_ENV
 
+      - uses: launchdarkly/gh-actions/actions/release-secrets@release-secrets-v1.0.1
+        name: Get secrets
+        with:
+          aws_assume_role: ${{ vars.AWS_ROLE_ARN }}
+
+      - name: Download snk
```
kinyoklion, line 32: The release-secrets actions supports S3. So it should be used directly.

kinyoklion (changes requested):

.github/workflows/publish-docs.yml
kinyoklion: https://github.com/launchdarkly/openfeature-java-server/blob/c6b2edff3f1b33cda0117aadde7aa52c67b61dbc/.github/workflows/publish.yml#L55

tanderson-ld: Done

kinyoklion (approved):

PR #12, opened 2024-08-21
fix: improved accuracy of stream interruption error message.

tanderson-ld:
Updated error message to not assume root cause is an outage.  Message matches other server SDKs that were also updated.

### DIFF
--- a/pkgs/sdk/server/src/Internal/DataSources/DataSourceOutageTracker.cs
+++ b/pkgs/sdk/server/src/Internal/DataSources/DataSourceOutageTracker.cs
@@ -96,7 +96,7 @@ namespace LaunchDarkly.Sdk.Server.Internal.DataSources
                     return;
                 }
                 var errorsDesc = string.Join(", ", _errorCounts.Select(kv => DescribeErrorCount(kv.Key, kv.Value)));
-                _log.Error("LaunchDarkly data source outage - updates have been unavailable for at least {0} with the following errors: {1}",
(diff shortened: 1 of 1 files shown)


pkgs/sdk/server/src/Internal/DataSources/DataSourceOutageTracker.cs
```diff
@@ -96,7 +96,7 @@ private async Task WaitForTimeout(Task outageEnded)
                     return;
                 }
                 var errorsDesc = string.Join(", ", _errorCounts.Select(kv => DescribeErrorCount(kv.Key, kv.Value)));
-                _log.Error("LaunchDarkly data source outage - updates have been unavailable for at least {0} with the following errors: {1}",
+                _log.Error("A streaming connection to LaunchDarkly has not been established within {0} after the connection was interrupted. The following errors were encountered: {1}",
```
kinyoklion, line 99: This probably only makes sense for streaming, but I do want to double check if this is streaming only. Otherwise the message could be interpolated.

tanderson-ld: I thought the same thing, it doesn't seem like it should be streaming specific.  Java currently says "streaming", so I just mimic'd that.  I'll also dig a bit.

tanderson-ld: Removed the data source type from the string.

kinyoklion (approved):

PR #19, opened 2024-10-22
feat: report prerequisite relations in AllFlagState

cwaldren-ld:
This PR updates `AllFlagsState` to track prerequisite evaluations. 

This didn't require modifying the Evaluator interface, as it already returns prerequisites as part of the `EvalResult`. 

When the returned `FlagState` is marshaled to JSON, it will now contain the prerequisite info for each flag.

### DIFF
--- a/pkgs/sdk/server/src/FeatureFlagsState.cs
+++ b/pkgs/sdk/server/src/FeatureFlagsState.cs
@@ -168,6 +168,18 @@ namespace LaunchDarkly.Sdk.Server
         /// </summary>
         /// <param name="flagKey">the flag key</param>
         /// <param name="result">the evaluation result</param>
+        /// <returns></returns>
+        public FeatureFlagsStateBuilder AddFlag(string flagKey, EvaluationDetail<LdValue> result)
+        {
+            return AddFlag(flagKey, result, new List<string>());
+        }
+
+
+        /// <summary>
+        /// Adds the result of a flag evaluation, including direct prerequisites.
+        /// </summary>
+        /// <param name="flagKey">the flag key</param>
+        /// <param name="result">the evaluation result</param>
         /// <param name="prerequisites">the direct prerequisites evaluated for this flag</param>
         /// <returns></returns>
         public FeatureFlagsStateBuilder AddFlag(string flagKey, EvaluationDetail<LdValue> result, List<string> prerequisites)
@@ -205,7 +217,7 @@ namespace LaunchDarkly.Sdk.Server
         }
     }
 
-    internal struct FlagState
+    internal struct FlagState : IEquatable<FlagState>
     {
         internal LdValue Value { get; set; }
         internal int? Variation { get; set; }
@@ -217,19 +229,29 @@ namespace LaunchDarkly.Sdk.Server
 
         internal List<string> Prerequisites { get; set; }
 
-        public override bool Equals(object other)
+        public bool Equals(FlagState o)
         {
-            if (other is FlagState o)
-            {
-                return Variation == o.Variation &&
-                       Version == o.Version &&
-                       TrackEvents == o.TrackEvents &&
-                       TrackReason == o.TrackReason &&
-                       DebugEventsUntilDate.Equals(o.DebugEventsUntilDate) &&
-                       Object.Equals(Reason, o.Reason) &&
-                       Prerequisites == o.Prerequisites;
-            }
-            return false;
+            return Variation == o.Variation &&
+                   Version == o.Version &&
+                   TrackEvents == o.TrackEvents &&
+                   TrackReason == o.TrackReason &&
+                   DebugEventsUntilDate.Equals(o.DebugEventsUntilDate) &&
+                   Object.Equals(Reason, o.Reason) &&
+                   Prerequisites.SequenceEqual(o.Prerequisites);
+        }
+        public override bool Equals(object obj)
+        {
+            return obj is FlagState other && Equals(other);
+        }
+
+        public static bool operator ==(FlagState lhs, FlagState rhs)
+        {
+            return lhs.Equals(rhs);
+        }
+
+        public static bool operator !=(FlagState lhs, FlagState rhs)
+        {
+            return !(lhs == rhs);
         }
 
         public override int GetHashCode()
@@ -310,7 +332,10 @@ namespace LaunchDarkly.Sdk.Server
                         for (var flagsObj = RequireObject(ref reader); flagsObj.Next(ref reader);)
                         {
                             var subKey = flagsObj.Name;
-                            var flag = flags.ContainsKey(subKey) ? flags[subKey] : new FlagState();
+                            var flag = flags.ContainsKey(subKey)
+                                ? flags[subKey]
+                                : new FlagState() { Prerequisites = new List<string>() };
+
                             for (var metaObj = RequireObject(ref reader); metaObj.Next(ref reader);)
                             {
                                 switch (metaObj.Name)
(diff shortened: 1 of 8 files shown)


pkgs/sdk/server/src/LdClient.cs
```diff
@@ -371,8 +372,7 @@ public FeatureFlagsState AllFlagsState(Context context, params FlagsStateOption[
 
             var builder = new FeatureFlagsStateBuilder(options);
```
cwaldren-ld, line 373: These weren't used at all in this function. They are used internal to the `FeatureFlagsStateBuilder`, presumably they didn't get cleaned up when that was introduced.

pkgs/sdk/server/src/LdClient.cs
```diff
@@ -397,6 +397,11 @@ public FeatureFlagsState AllFlagsState(Context context, params FlagsStateOption[
                 {
                     EvaluatorTypes.EvalResult result = _evaluator.Evaluate(flag, context);
                     bool inExperiment = EventFactory.IsExperiment(flag, result.Result.Reason);
+
+                    var directPrerequisites = result.PrerequisiteEvals.Where(
+                        e => e.FlagKey == flag.Key)
+                        .Select(p => p.PrerequisiteFlag.Key).ToList();
```
cwaldren-ld, line 403: Converting to a list here, rather than leaving as an `IEnumerable`, because the intent is that order is preserved.

pkgs/sdk/server/src/FeatureFlagsState.cs
```diff
@@ -295,7 +332,10 @@ public override FeatureFlagsState Read(ref Utf8JsonReader reader, Type typeToCon
                         for (var flagsObj = RequireObject(ref reader); flagsObj.Next(ref reader);)
                         {
                             var subKey = flagsObj.Name;
-                            var flag = flags.ContainsKey(subKey) ? flags[subKey] : new FlagState();
```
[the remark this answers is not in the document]
tanderson-ld: I agree with the constructor idea or something similar.  Faced a similar issue in Java wrt to nullity, but luckily the array constructor with capacity 0 uses a static instance under the hood for performance.

pkgs/sdk/server/src/FeatureFlagsState.cs
```diff
@@ -271,6 +300,14 @@ public override void Write(Utf8JsonWriter w, FeatureFlagsState state, JsonSerial
                     w.WritePropertyName("reason");
                     EvaluationReasonConverter.WriteJsonValue(meta.Reason.Value, w);
                 }
+                if (meta.Prerequisites.Count > 0) {
```
tanderson-ld, line 303: Do you need a null check here?

tanderson-ld (approved):

pkgs/sdk/server/src/FeatureFlagsState.cs
```diff
@@ -169,6 +170,19 @@ public FeatureFlagsStateBuilder Valid(bool valid)
...
+        }
+
+
+        /// <summary>
+        /// Adds the result of a flag evaluation, including direct prerequisites.
+        /// </summary>
+        /// <param name="flagKey">the flag key</param>
+        /// <param name="result">the evaluation result</param>
+        /// <param name="prerequisites">the direct prerequisites evaluated for this flag</param>
+        /// <returns></returns>
```
kinyoklion, line 184: Empty tag.

pkgs/sdk/server/src/FeatureFlagsState.cs
```diff
@@ -194,14 +209,15 @@ internal FeatureFlagsStateBuilder AddFlag(string flagKey, LdValue value, int? va
...
+                TrackReason = trackReason,
+                Prerequisites = prerequisites
             };
             _flags[flagKey] = flag;
             return this;
         }
     }
 
-    internal struct FlagState
+    internal struct FlagState : IEquatable<FlagState>
```
kinyoklion, line 220: Non-nullable prerequisites feel a bit strange to me.

pkgs/sdk/server/src/FeatureFlagsState.cs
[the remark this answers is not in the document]
kinyoklion: Theoretically we should have struct constructors, but not parameter-less constructors. Ideally we would be creating a fully populated readonly struct instead of poking stuff into it. (containing an immutable list of prerequisites, or null)

pkgs/sdk/server/src/FeatureFlagsState.cs
```diff
@@ -295,7 +332,15 @@ public override FeatureFlagsState Read(ref Utf8JsonReader reader, Type typeToCon
...
                         {
                             var subKey = flagsObj.Name;
-                            var flag = flags.ContainsKey(subKey) ? flags[subKey] : new FlagState();
+
+                            var flag = flags.ContainsKey(subKey)
+                                ? flags[subKey]
+                                : new FlagState
+                                {
+                                    // Most flags have no prerequisites, don't allocate unless we need to.
+                                    Prerequisites = new List<string>(0)
```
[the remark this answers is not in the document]
cwaldren-ld: Oops, I intended for my comment to say "don't allocate [capacity] unless we need to". I wanted to have the semantics of "you can iterate this list and never worry about it being null" from the consumer's perspective.

pkgs/sdk/server/src/FeatureFlagsState.cs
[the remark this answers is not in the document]
cwaldren-ld: With the way we are deserializing JSON, we'd need to have some kind of builder pattern (or temp variables) to achieve that (creating the readonly struct in one go.)

pkgs/sdk/server/src/FeatureFlagsState.cs
cwaldren-ld: Fair. Is that from a "what a library consumer would expect in C#" kind of way, or a "how structs work" way?

pkgs/sdk/server/src/FeatureFlagsState.cs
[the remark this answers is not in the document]
kinyoklion: The contract here is internal to us (internal struct, and the public flag state only has specific accessors), so the non-garbage generating contract seems better. Or a shared empty list. Though ideally we would make the contact of the struct immutable.

pkgs/sdk/server/src/FeatureFlagsState.cs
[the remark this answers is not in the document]
kinyoklion: If the temp variables are references, then they aren't really temps, the struct holds a reference. Though it maybe it problematic to adjust. It isn't at an assignment level though. Assign to a shared list and then assign to the populated list.

pkgs/sdk/server/src/FeatureFlagsState.cs
kinyoklion: So, it feels strange in the language to me. We have null-state analysis, so we should be able to safely mark things null-able and use them. https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references#null-state-analysis

From a struct perspective, ideally structs only contain other value types, but that isn't always possible. Next is for structs to only contain immutable reference types in addition to value types. Least favorable is this mix, but also not always avoidable. A struct with mostly not references, with value semantics, is better than a class here probably. But it doesn't behave like a struct. In that if you alias it and mutate the list, then both lists mutate.

kinyoklion: If we make an immutable list, and then assign the entire list, then it at least makes it more toward the normal side. It could also be exposed as an `IReadOnlyList<T>`.

kinyoklion: I don't think this is blocking though, as it isn't part of the external API.

kinyoklion (approved):

kinyoklion (approved):

PR #25, opened 2024-10-24
build: enable client SDK unit tests in CI

cwaldren-ld:


.github/actions/ci/action.yml
```diff
@@ -7,6 +7,14 @@ inputs:
   test_project_file:
     description: 'Path to the test project file.  If provided, run unit tests, otherwise skip them.'
     required: false
+  target_test_framework:
+    description: 'The target test framework to use.'
+    required: false
+    default: 'net6.0'
+  test_verbosity:
```
cwaldren-ld, line 14: Without this, it just appears like the tests are hanging in CI. I can see the value of not having this when running locally, but it's nice to have the record when waiting remotely.

.github/actions/ci/action.yml
```diff
@@ -35,4 +43,4 @@ runs:
       shell: bash
       run: |
         dotnet restore ${{ inputs.test_project_file }}
-        dotnet test --framework=net6.0 ${{ inputs.test_project_file }}
+        dotnet test -v=${{ inputs.test_verbosity }} --framework=${{ inputs.target_test_framework }} ${{ inputs.test_project_file }}
```
cwaldren-ld, line 46: Previously, it was passing net6 as the framework. I believe this was causing it to silently just not run tests at all. The target for the client tests is net7.

tanderson-ld (approved):

PR #36, opened 2024-11-05
refactor: implement AI Client

cwaldren-ld:
This implements the guts of the AI client, including fetching Model configuration, interpolation, and tests.


pkgs/sdk/server-ai/src/LdAiClient.cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            _client.GetLogger().Error("Unable to parse AI model config for key {0}: {1}", key, e.Message);
+            return null;
+        }
+    }
+
+    /// <summary>
+    /// THD
+    /// </summary>
+    public void Dispose()
```
cwaldren-ld, line 168: Unclear if this should really be implemented or not. If you are instantiating the client directly with a server client, then this makes sense.

If you are just wrapping it, and might use the base client for other stuff, then no.

pkgs/sdk/server-ai/src/LdAiClient.cs
```diff
@@ -0,0 +1,169 @@
...
+    /// <param name="defaultValue">the default config, if unable to retrieve from LaunchDarkly</param>
+    /// <param name="variables">the list of variables used when interpolating the prompt</param>
+    /// <returns>an AI config tracker</returns>
+    public LdAiConfigTracker ModelConfig(string key, Context context, LdAiConfig defaultValue,
+        IReadOnlyDictionary<string, object> variables = null)
+    {
+
+        var result = _client.JsonVariation(key, context, LdValue.Null);
+
+        var defaultTracker = new LdAiConfigTracker(_client, key, defaultValue, context);
```
cwaldren-ld, line 68: I was previously using a detailed evaluation here, which had the ability to directly confirm if the default was returned. This was nice. 

Using a normal eval, I can only detect that by passing in a particular default and confirming that is what came back out. I could pass in some sentinel value that is highly unlikely to be anything else.

This is in service of not having to serialize the passed in `LdAiConfig` and pass it into the SDK.

pkgs/sdk/server-ai/README.md
```diff
@@ -22,7 +22,7 @@ The .NET build tools should automatically load the most appropriate build of the
 
 ## Getting started
 
-Refer to the [SDK documentation](https://docs.launchdarkly.com/sdk/server-side/dotnet#getting-started) for instructions on getting started with using the SDK.
+Refer to the [SDK documentation](TODO) for instructions on getting started with using the SDK.
```
cwaldren-ld, line 25: TODO: update when available

pkgs/sdk/server-ai/src/Config/LdAiConfig.cs
```diff
@@ -0,0 +1,159 @@
...
+    /// <summary>
+    /// The model parameters associated with the config.
+    /// </summary>
+    public readonly IReadOnlyDictionary<string, object> Model;
+
+
+
+    internal LdAiConfig(bool enabled, IEnumerable<Message> prompt, Meta meta, IReadOnlyDictionary<string, object> model)
+    {
+        Model = model ?? new Dictionary<string, object>();
```
kinyoklion, line 128: Should `object` here be LDValue? For JSON production purposes, no cycles, etc? (Maybe I will learn more as I proceed.)

pkgs/sdk/server-ai/src/Config/LdAiConfig.cs
```diff
@@ -0,0 +1,159 @@
...
+        VersionKey = meta?.VersionKey ?? "";
+        Enabled = enabled;
+    }
+
+
+    /// <summary>
+    /// Creates a new LdAiConfig builder.
+    /// </summary>
+    /// <returns>a new builder</returns>
+    public static Builder New() => new();
```
kinyoklion, line 139: I think that the constructor should be internal if we are going to do this.

pkgs/sdk/server-ai/src/DataModel/DataModel.cs
```diff
@@ -0,0 +1,89 @@
...
+    /// <summary>
+    /// The version key.
+    /// </summary>
+    [JsonPropertyName("versionKey")]
+    public string VersionKey { get; set; }
+
+    /// <summary>
+    /// If the config is enabled.
+    /// </summary>
+
```
kinyoklion, line 39: Few strange blank lines in this file.

pkgs/sdk/server-ai/src/DataModel/DataModel.cs
```diff
@@ -0,0 +1,89 @@
+using System.Collections.Generic;
```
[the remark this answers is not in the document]
kinyoklion: You should be able to tell it to use a parameterized constructor `JsonContructor` which allows creating immutable types, but you would have to have a full parameterized constructor for the records (because no `init`).

pkgs/sdk/server-ai/src/Interfaces/ILaunchDarklyClient.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// <param name="data">associated data</param>
+    /// <param name="metricValue">metric value</param>
+    void Track(string name, Context context, LdValue data, double metricValue);
+
+    /// <summary>
+    /// Returns a logger.
+    /// </summary>
+    /// <returns>a logger</returns>
+    ILogger GetLogger();
+
```
kinyoklion, line 32: Unexpected blank.

pkgs/sdk/server-ai/src/Interfaces/ILogger.cs
```diff
@@ -0,0 +1,22 @@
+namespace LaunchDarkly.Sdk.Server.Ai.Interfaces;
+
+/// <summary>
+/// Log interface required by the AI Client.
+/// </summary>
+public interface ILogger
+{
+
```
kinyoklion, line 8:
```suggestion
```

pkgs/sdk/server-ai/src/LdAiClient.cs
```diff
@@ -0,0 +1,169 @@
...
+using LaunchDarkly.Sdk.Server.Interfaces;
+using Mustache;
+
+namespace LaunchDarkly.Sdk.Server.Ai;
+
+/// <summary>
+/// The LaunchDarkly AI client. The client is capable of retrieving AI configurations from LaunchDarkly,
+/// and generating events specific to usage of the AI configuration when interacting with model providers.
+/// </summary>
+public sealed class LdAiClient
```
kinyoklion, line 19: Typically we should have an interface on this so the customer can mock it. (And all the compatibility inconveniences with that.)

cwaldren-ld: Added an `ILdAiClient` and also changed the ModelConfig return to an `ILdAiConfigTracker`.

cwaldren-ld:
Updated to pass `defaultValue.ToLdValue()` directly into the underlying SDK. This recursively converts the existing `LdAiConfig` into an `LdValue`.

pkgs/sdk/server-ai/README.md
mmrj: The link will be `https://docs.launchdarkly.com/sdk/ai/dotnet` but this PR will probably end up merging before the docs one. So you could also omit until after the docs are published if you want (I believe that's what Ryan did for Node).

kinyoklion (approved):

PR #43, opened 2024-11-13
fix: catch exceptions thrown by template interpolation

cwaldren-ld:
This is a first pass at catching exceptions generated by the interpolation process. It wraps each call to interpolate with a try/catch and individually reports an error via the logger.

I know that the exception message in this particular case (in unit test) is pretty useless, but it also seems bad to simply swallow the error with no extra context.

### DIFF
--- a/pkgs/sdk/server-ai/src/LdAiClient.cs
+++ b/pkgs/sdk/server-ai/src/LdAiClient.cs
@@ -86,7 +86,8 @@ public sealed class LdAiClient : ILdAiClient
                 }
                 catch (Exception ex)
                 {
-                    _logger.Error($"AI model config prompt has malformed message at index {i}: {ex.Message} (returning default config)");
+                    _logger.Error(
+                        $"AI model config prompt has malformed message at index {i}: {ex.Message} (returning default config, which will not contain interpolated prompt messages)");
                     return new LdAiConfigTracker(_client, key, defaultValue, context);
                 }
             }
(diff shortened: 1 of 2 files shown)


pkgs/sdk/server-ai/test/InterpolationTests.cs
```diff
@@ -117,6 +118,35 @@ public void TestInterpolationWithArraySectionWorks()
         Assert.Equal("hello world ! ", result);
     }
 
+    [Fact]
+    public void TestInterpolationMalformed()
```
cwaldren-ld, line 122: Not the most comprehensive of test suites. I'll add more to this.

pkgs/sdk/server-ai/src/LdAiClient.cs
```diff
@@ -73,10 +73,27 @@ public ILdAiConfigTracker ModelConfig(string key, Context context, LdAiConfig de
...
+            for (var i = 0; i < parsed.Prompt.Count; i++)
+            {
+                try
+                {
+                    var content = InterpolateTemplate(parsed.Prompt[i].Content, mergedVariables);
+                    prompt.Add(new LdAiConfig.Message(content, parsed.Prompt[i].Role));
+                }
+                catch (Exception ex)
+                {
+                    _logger.Error($"AI model config prompt has malformed message at index {i}: {ex.Message} (returning default config)");
```
kinyoklion, line 89: I think we should extend the message with a warning about the interpolation. "The returned configuration will not have interpolated prompts." Or something that indicates that not only is it the default, but it is not interpolated.

Because typically the default will be interpolated (because it goes into and out of variation).

cwaldren-ld: Ah, that's a good point.

kinyoklion (approved):

PR #52, opened 2024-11-22
fix: rename ModelConfig method to Config

cwaldren-ld:
The latest specification renames `ModelConfig` to `Config`.


mmrj (approved):
thank you!

PR #54, opened 2024-11-23
fix: add setter for ModelId

cwaldren-ld:


mmrj (approved):
thank you! changes make sense to me but should probably have an actual Eng review too : )

tanderson-ld (approved):

PR #61, opened 2024-12-09
refactor!: rename model.id and provider.id to name

cwaldren-ld:


kinyoklion (approved):

mmrj (commented):
Should `SetModelProviderId` (line 178) and `SetModelId` (line 167) keep their current names? I was expecting them to change but may not be understanding the proposed updates correctly.

cwaldren-ld:
@mmrj  good eye, you are correct. I've updated.

PR #66, opened 2024-12-17
chore: Expand doc for exception handling

keelerm84:


pkgs/sdk/server-ai/src/Interfaces/ILdAiConfigTracker.cs
```diff
@@ -47,13 +51,25 @@ public interface ILdAiConfigTracker
 
     /// <summary>
     /// Tracks an unsuccessful generation event related to this config.
+    ///
+    ///
```
kinyoklion, line 55: Intentional?

kinyoklion (approved):

pkgs/sdk/server-ai/src/Interfaces/ILdAiConfigTracker.cs
keelerm84: Nope

PR #67, opened 2025-01-23
track TimeToFirstToken in LdAiConfigTracker

moribellamy:


pkgs/sdk/server-ai/src/Interfaces/ILdAiConfigTracker.cs
```diff
@@ -36,6 +36,11 @@ public interface ILdAiConfigTracker
     /// <typeparam name="T">type of the task's result</typeparam>
     /// <returns>the task</returns>
     public Task<T> TrackDurationOfTask<T>(Task<T> task);
+    
+    /// <summary>
+    /// Tracks the time it takes for the first token to be generated.
+    /// </summary>
```
keelerm84, line 42:
```suggestion
    /// </summary>
    /// <param name="timeToFirstTokenMs">the duration in milliseconds</param>
```

tanderson-ld (approved):
I suspect the failure in the tests is due to microsoft releasing the next version of the workloads we use.  I can try to dig into it as I was the last to touch that layer of this repo.  What is your time frame on this change?

moribellamy:
> I suspect the failure in the tests is due to microsoft releasing the next version of the workloads we use. I can try to dig into it as I was the last to touch that layer of this repo. What is your time frame on this change?

can we shoot for early next week?

should i hold of on submitting until the CI is fixed?

tanderson-ld:
Yeah I can try to take a look early next week.  We do need to address them before this merges to help maintain the CI/CD

tanderson-ld:
Taking a look today.

PR #73, opened 2025-02-04
ci: Fix build error

keelerm84:


pkgs/sdk/client/src/LaunchDarkly.ClientSdk.csproj
```diff
@@ -33,6 +33,7 @@
         <SymbolPackageFormat>snupkg</SymbolPackageFormat>
         <RootNamespace>LaunchDarkly.Sdk.Client</RootNamespace>
         <IncludeSourceRevisionInInformationalVersion>false</IncludeSourceRevisionInInformationalVersion>
+        <CheckEolWorkloads>false</CheckEolWorkloads>
```
kinyoklion, line 36: I think maybe @tanderson-ld  was looking at updating the workloads? It would be better to not be using EOL workloads.

kinyoklion (approved):

PR #78, opened 2025-02-27
POC of an approach to adding async

jackmott:
**Requirements**

- [ ] I have added test coverage for new or changed functionality
- [ ] I have followed the repository's [pull request submission guidelines](../blob/main/CONTRIBUTING.md#submitting-pull-requests)

**Related issues**

A proof of concept for adding async to the client, if LD is happy with this approach I can complete it and add test coverage. The intent here it to be backwards compatible and follow .NET conventions by adding Async versions of the calls.

Provide a clear and concise description of any alternative solutions or features you've considered.

**Additional context**

Add any other context about the pull request here.

### DIFF
--- a/pkgs/sdk/server/src/Internal/DataStores/InMemoryDataStore.cs
+++ b/pkgs/sdk/server/src/Internal/DataStores/InMemoryDataStore.cs
@@ -1,4 +1,6 @@
 ﻿using System.Collections.Immutable;
+using System.Threading;
+using System.Threading.Tasks;
 using LaunchDarkly.Sdk.Server.Subsystems;
 
 using static LaunchDarkly.Sdk.Server.Subsystems.DataStoreTypes;
@@ -58,6 +60,19 @@ namespace LaunchDarkly.Sdk.Server.Internal.DataStores
                 return null;
             }
             return item;
+        }
+
+        public Task<ItemDescriptor?> GetAsync(DataKind kind, string key, CancellationToken cancelationToken = default)
+        {
+            if (!Items.TryGetValue(kind, out var itemsOfKind))
+            {
+                return null;
+            }
+            if (!itemsOfKind.TryGetValue(key, out var item))
+            {
+                return null;
+            }
+            return Task.FromResult<ItemDescriptor?>(item);
         }
 
         public KeyedItems<ItemDescriptor> GetAll(DataKind kind)
--- a/pkgs/sdk/server/src/Internal/DataStores/PersistentStoreWrapper.cs
+++ b/pkgs/sdk/server/src/Internal/DataStores/PersistentStoreWrapper.cs
@@ -2,6 +2,8 @@
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 using LaunchDarkly.Cache;
 using LaunchDarkly.Logging;
 using LaunchDarkly.Sdk.Internal;
@@ -185,6 +187,22 @@ namespace LaunchDarkly.Sdk.Server.Internal.DataStores
             }
         }
 
+        public async Task<ItemDescriptor?> GetAsync(DataKind kind, string key, CancellationToken cancellationToken = default)
+        {
+            try
+            {
+                var ret = _itemCache is null ? await GetAndDeserializeItemAsync(kind, key) :
+                    _itemCache.Get(new CacheKey(kind, key));
+                ProcessError(null);
+                return ret;
+            }
+            catch (Exception e)
+            {
+                ProcessError(e);
+                throw;
+            }
+        }
+
         public KeyedItems<ItemDescriptor> GetAll(DataKind kind)
         {
             try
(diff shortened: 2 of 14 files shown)


pkgs/sdk/server/src/LdClient.cs
```diff
@@ -658,6 +779,16 @@ private FeatureFlag GetFlag(string key)
             return null;
         }
 
+        private async Task<FeatureFlag> GetFlagAsync(string key, CancellationToken cancellationToken = default)
+        {
+            var maybeItem = await _dataStore.GetAsync(DataModel.Features, key,cancellationToken);
```
dahlbyk, line 784: We'd definitely want to use `ConfigureAwait(false)` on all of these.

pkgs/sdk/server/src/Internal/DataStores/PersistentStoreWrapper.cs
```diff
@@ -185,6 +187,22 @@ public void Init(FullDataSet<ItemDescriptor> items)
             }
         }
 
+        public async Task<ItemDescriptor?> GetAsync(DataKind kind, string key, CancellationToken cancellationToken = default)
+        {
+            try
+            {
+                var ret = _itemCache is null ? await GetAndDeserializeItemAsync(kind, key, cancellationToken) :
+                    _itemCache.Get(new CacheKey(kind, key));
```
dahlbyk, line 195: In high-volume uses we would mostly expect sync cache hits, which makes this a good candidate for `ValueTask<>`.

dahlbyk (commented):
Would love to see this!

pkgs/sdk/server/src/Internal/DataStores/PersistentStoreWrapper.cs
```diff
@@ -185,6 +187,22 @@ public void Init(FullDataSet<ItemDescriptor> items)
             }
         }
 
+        public async ValueTask<ItemDescriptor?> GetAsync(DataKind kind, string key, CancellationToken cancellationToken = default)
+        {
+            try
+            {
+                var ret = _itemCache is null ? await GetAndDeserializeItemAsync(kind, key, cancellationToken).ConfigureAwait(false) :
+                    _itemCache.Get(new CacheKey(kind, key));
```
dahlbyk, line 195: So I was going to say that since the `_itemCache` branch here is all sync there would probably be a perf benefit to splitting into paths with and without `async`/`await`.

But that made me realize that we don't _want_ an all sync path. The cache [loader](https://github.com/launchdarkly/dotnet-core/blob/2dc00b4aac85f8c36310031b0ea1283a6e56009f/pkgs/sdk/server/src/Internal/DataStores/PersistentStoreWrapper.cs#L68-L69) is [async over sync](https://github.com/launchdarkly/dotnet-core/blob/2dc00b4aac85f8c36310031b0ea1283a6e56009f/pkgs/sdk/server/src/Internal/DataStores/PersistentStoreWrapper.cs#L328-L329), so we haven't actually avoided any blocking unless we disable the cache. 😭 We'd need to update [LaunchDarkly.Cache](https://github.com/launchdarkly/dotnet-cache/) with `IAsyncCache`/`IAsyncSingleValueCache` and wire it through.

dahlbyk (commented):
See last comment. 😢

pkgs/sdk/server/src/Internal/DataStores/InMemoryDataStore.cs
```diff
@@ -60,6 +62,23 @@ public void Init(FullDataSet<ItemDescriptor> data)
...
+            }
+            if (!itemsOfKind.TryGetValue(key, out var item))
+            {
+                return new ValueTask<ItemDescriptor?>();
+            }
+#if NET31_OR_GREATER
+            return ValueTask.FromResult<ItemDescriptor?>(item);
+#else
+            return new ValueTask<ItemDescriptor?>(Task.FromResult<ItemDescriptor?>(item));
+#endif
```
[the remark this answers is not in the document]
jackmott: that's not a thing in net 46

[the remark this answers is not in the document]
dahlbyk: Add conditional reference to https://www.nuget.org/packages/System.Threading.Tasks.Extensions for `net462`?

louis-launchdarkly:
Hello @jackmott and @dahlbyk, thank you so much for the contribution. While the SDK team cannot get to this immediately, we will look at this contribution, as we want to make the SDK fit the platform's use cases.

To understand the platform better, is async usage more common in .NET server applications these days? Is there a concern that the persistent store data read slows down the SDK's performance?

Filed Internally as SDK-1096

jackmott:
Hello @louis-launchdarkly !

Yes async is now very much the default approach for .NET server application.  For some reading on this you can refer to:  https://learn.microsoft.com/en-us/aspnet/core/fundamentals/best-practices?view=aspnetcore-9.0#avoid-blocking-calls

While we understand the past position of LaunchDarkly is that caching makes blocking rare enough to mitigate this issue, it would still be better for this to be a fully async path, and we believe ValueTask in particular can make for an efficient async option here.

We have a lot of expertise in  house at Olo making our own legacy code async and so we may be able to help work through any tricky issues.

arielnahom:
Would be great to have async support! Launch darkly team - Please add it.

PR #81, opened 2025-03-04
feat: Add environment ID support for hooks.

kinyoklion:
This PR pipes the support for the environment ID header from the event source, through the data sources, and into the data store, and then exposes it to the evaluation series hook context.

Each step can optionally support the environment ID/headers for compatibility.

This PR will be held for back-end support and testing.

### DIFF
--- a/pkgs/sdk/server/src/Internal/DataSources/DataSourceUpdatesImpl.cs
+++ b/pkgs/sdk/server/src/Internal/DataSources/DataSourceUpdatesImpl.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Threading.Tasks;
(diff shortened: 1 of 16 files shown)


pkgs/sdk/server/src/Internal/DataSources/DataSourceUpdatesImpl.cs
```diff
@@ -84,7 +85,7 @@ internal DataSourceUpdatesImpl(
                 StateSince = DateTime.Now,
                 LastError = null
             };
-            _status = new StateMonitor<DataSourceStatus, StateAndError>(initialStatus, MaybeUpdateStatus, _log);            
+            _status = new StateMonitor<DataSourceStatus, StateAndError>(initialStatus, MaybeUpdateStatus, _log);
```
kinyoklion, line 88: Just a whitespace change.

pkgs/sdk/server/src/Internal/DataSources/DataSourceUpdatesImpl.cs
```diff
@@ -326,5 +287,67 @@ private void ReportStoreFailure(Exception e)
         }
 
         #endregion
+
+       #region IDataSourceUpdatesHeaders methods
+        public bool InitWithHeaders(FullDataSet<ItemDescriptor> allData, IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers)
+        {
+            ImmutableDictionary<DataKind, ImmutableDictionary<string, ItemDescriptor>> oldData = null;
```
kinyoklion, line 294: Most of the method body is from the original init.

tanderson-ld (approved):

PR #82, opened 2025-03-04
feat: Add environment id support for the OTEL hook.

kinyoklion:
This PR will need to be merged after the base PR, and the version required by the hook will need updated.

### DIFF
--- a/pkgs/telemetry/src/LaunchDarkly.ServerSdk.Telemetry.csproj
+++ b/pkgs/telemetry/src/LaunchDarkly.ServerSdk.Telemetry.csproj
@@ -37,7 +37,7 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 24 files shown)


pkgs/telemetry/src/LaunchDarkly.ServerSdk.Telemetry.csproj
```diff
@@ -37,7 +37,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="LaunchDarkly.ServerSdk" Version="8.*" />
+    <PackageReference Include="LaunchDarkly.ServerSdk" Version="[8.7, 9.0)" />
```
kinyoklion, line 40: Hook 1.1.0 will require SDK 8.7.0 or greater. I constrained to less than 9 for now.

tanderson-ld (approved):

PR #98, opened 2025-05-08
chore: Fix release and CI process for the ClientSDK

jsonbailey:
We cannot build the client with .NET8 or higher with the maui workloads and still target net6. (https://learn.microsoft.com/en-us/dotnet/core/tools/sdk-errors/netsdk1202) Enabled the global.json during the release build so the dotnet version is consistent with CI.


tanderson-ld (approved):

jsonbailey:
I pushed this branch mainly as a proof of concept. I'm not sure if we want to enforce the global.json file for local development or not but it does look like it needs to be enforced for at least the CI action (this PR removes that but its the current behavior on main) and the build-release action which doesn't support it and caused a failure when releasing the ClientSDK.

Issue #89, opened 2025-04-03
LaunchDarkly.ClientSDK 5.3.0 throws `Format_InvalidStringWithOffsetAndReason`


ryanalford-suzy:
**Is this a support request?**
We are seeing an increasing number of exceptions being thrown in the ClientSDK for .Net/Maui. We are only seeing it in release builds, but I can't totally rule out that it would happen in debug builds. We just haven't seen it in debug builds.

It seems like the exception is happening in the LaunchDarkly.InternalSDK. 

**Describe the bug**
See screenshots below

**To reproduce**
N/A

**Expected behavior**
Should not throw exceptions?

**Logs**
See screenshots below

**SDK version**
LaunchDarkly.ClientSDK 5.3.0

**Language version, developer tools**
.Net 8.0, Maui 8.0.100

**OS/platform**
Only seems to happen on iOS. I've got iOS reports on iOS 15.5 all the way through 18.3. 

I went through our Android crash/error reporting and didn't find any reports.

**Additional context**
Since I can't give you access to the crash report, I can try to give screenshots of the stack trace and error message.

![Image](https://github.com/user-attachments/assets/539d7d0e-3c98-4a5f-a16d-dc52a155265a)

ryanalford-suzy:
Over the last 24 hours, 104 users had it occur 148 times. Over the last 7 days, 708 users had it occur ~4200 times.

![Image](https://github.com/user-attachments/assets/cf8accc3-81e7-49ea-a74f-62f8b469a4a7)

![Image](https://github.com/user-attachments/assets/70bdb14b-56ff-4078-8514-f3fd541cfd96)

tanderson-ld:
Thank you for the quick response. Did anything in the project change recently? SDK Version update, target .NET version, etc...?

ryanalford-suzy:
We updated SDK from version 5.2.1 to 5.3.0 about 3 months ago(we are on our third release since then). At that time, we also started adding a custom attribute for the user Context.

![Image](https://github.com/user-attachments/assets/9bab6fc1-377d-4234-b310-e59d78b52d79)

Pretty straight forward code change.

Our latest version is the first time we've had Sentry crash reporting implemented. We were using AppCenter before. I went through AppCenter's crash reports and didn't see this issue. I only started seeing it after moving to Sentry. Not sure if AppCenter was just missing it or not.

tanderson-ld:
[...]

PR #153, opened 2025-08-21
fix: Handling exceptions that occur during string.Format in the iOS ChannelImpl Logs

justinwojo:
We have multiple MAUI and .Net for iOS/Android based apps using the LaunchDarkly SDK, and have ran into thousands of crashes from this SDK due to a string.Format exception occurring in the Log methods of ChannelImpl on iOS.

We don't log anything to LaunchDarkly, so it seems logical that it's something internal that LaunchDarkly is logging.

This is a simple fix that prevents the crash on our apps, while providing LaunchDarkly with information relevant to fix the underlying format issue (if it's internal).

It's a non-breaking change.

**Requirements**

- [ ] I have added test coverage for new or changed functionality
- [X] I have followed the repository's [pull request submission guidelines](../blob/main/CONTRIBUTING.md#submitting-pull-requests)
- [X] I have validated my changes against all supported platform versions

**Related issues**

[Issue 89](https://github.com/launchdarkly/dotnet-core/issues/89)

In the case where string.Format would throw an exception, it'll gracefully fall back to a log event showing the provided format and params for further diagnosis.

I cannot find any alternatives without understanding what log event is causing the exception internally in the SDK.

**Additional context**

image

We only see this issue on iOS, it does not happen on Android.

### DIFF
--- a/pkgs/sdk/client/src/PlatformSpecific/Logging.ios.cs
+++ b/pkgs/sdk/client/src/PlatformSpecific/Logging.ios.cs
@@ -70,14 +70,45 @@ namespace LaunchDarkly.Sdk.Client.PlatformSpecific
                 public void Log(LogLevel level, object message) =>
                     LogString(level, message.ToString());
 
-                public void Log(LogLevel level, string format, object param) =>
-                    LogString(level, string.Format(format, param));
+                public void Log(LogLevel level, string format, object param)
+                {
+                    try
+                    {
+                        LogString(level, string.Format(format, param));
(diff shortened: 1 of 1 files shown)


jsonbailey:
@justinwojo thank you for the PR. After reviewing it and the original issue I think I have identified the root cause and am working to release an update to address the issue. If it does not resolve the issue then I will review this approach. I'm leaving it open until you can verify if the root fix addresses the issue for you. I will let you know once it is published.

PR #162, opened 2025-09-26
fix: Validate string format of SDK key

jsonbailey:


pkgs/sdk/server/src/Configuration.cs
```diff
@@ -187,7 +187,7 @@ internal Configuration(ConfigurationBuilder builder)
             Events = builder._events;
             Http = builder._http;
             Logging = builder._logging;
-            Offline = builder._offline;
+            Offline = builder._sdkKey == null || builder._offline;
```
jsonbailey, line 190: Do we want to force offline mode here? It seems unnecessary to attempt connecting if we know the key isn't valid.

pkgs/sdk/server/src/LdClient.cs
```diff
@@ -136,6 +136,12 @@ public LdClient(Configuration config)
             _log = logConfig.LogAdapter.Logger(logConfig.BaseLoggerName ?? LogNames.DefaultBase);
             _log.Info("Starting LaunchDarkly client {0}",
                 AssemblyVersions.GetAssemblyVersionStringForType(typeof(LdClient)));
+
+            if (_configuration.SdkKey == null)
```
kinyoklion, line 140: So, this is where I would expect `if (_configuration.SdkKey == null && !_configuration.offline)`, but it doesn't hurt to make them provide an empty string instead in that case. If we used nullable reference types they probably couldn't pass null.

kinyoklion: Oh, but they cannot set an empty string with the other validation. I fear this change probably would log errors in existing offline usage.

pkgs/shared/common/src/Helpers/ValidationUtils.cs
```diff
@@ -10,6 +10,33 @@ public static class ValidationUtils
...
+        public static string ValidateSdkKeyFormat(string sdkKey)
+        {
+
+            // For offline mode, we allow a null or empty SDK key and it is not invalid.
+            if (string.IsNullOrEmpty(sdkKey))
+            {
+                return null;
+            }
+
+            if (sdkKey.Length > 8192)
```
kinyoklion, line 27: I would move this to a constant.

pkgs/shared/common/src/Helpers/ValidationUtils.cs
```diff
@@ -10,6 +10,33 @@ public static class ValidationUtils
...
+            {
+                return "SDK key cannot be longer than 1024 characters.";
```
kinyoklion, line 29: And I would string interpolate that constant here.

pkgs/shared/common/src/Helpers/ValidationUtils.cs
```diff
@@ -9,9 +9,39 @@ namespace LaunchDarkly.Sdk.Helpers
...
+        /// </summary>
+        /// <param name="sdkKey">the SDK key to validate.</param>
+        /// <returns>Null if the input is valid, otherwise an error string describing the issue.</returns>
+        public static bool IsValidSdkKeyFormat(string sdkKey, out string errorMessage)
+        {
+
+            // For offline mode, we allow a null or empty SDK key and it is not invalid.
+            if (string.IsNullOrEmpty(sdkKey))
+            {
+                return true;
```
kinyoklion, line 25: I think we need to assign errorMessage to null.

kinyoklion (approved):

PR #167, opened 2025-10-03
feat: add all available attributes to a span

beeme1mr:
**Requirements**

- [x] I have added test coverage for new or changed functionality
- [x] I have validated my changes against all supported platform versions

**Related issues**

N/A: I can create one if necessary

This PR adds the ability to capture all available attributes when in "span mode". It will only add the additional attributes when `IncludeAllAttributesInSpan` is `true`; otherwise, the existing behavior is retained.

This change is useful because it's easier to work with spans than span events in some observability tools. With this change, spans can have the same attributes as span events.

We could leave it as is but it would make spans less useful to consumers.

**Additional context**

I talked to @kinyoklion about this on Slack about this change prior to opening the PR.

### DIFF
--- a/pkgs/telemetry/src/TracingHook.cs
+++ b/pkgs/telemetry/src/TracingHook.cs
@@ -16,12 +16,15 @@ namespace LaunchDarkly.Sdk.Server.Telemetry
     {
         private bool _createActivities;
         private bool _includeValue;
+
+        private bool _includeAllAttributesInSpan;
         private string _environmentId;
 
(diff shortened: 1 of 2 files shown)


beeme1mr:
Unfortunately, I don't have a way to test macOS. @kinyoklion, do you have a way to troubleshoot the failed test?

kinyoklion:
@beeme1mr It is an unrelated flake, so no concerns. We will be looking at it.

kinyoklion (approved):

PR #193, opened 2025-12-09
chore: Add data system configuration.

kinyoklion:
Adds a configuration for the data system, this configuration is not yet in use.

All methods, that will be eventually exposed, are currently set to internal.

---

> [!NOTE]
> Introduce an internal data system configuration with builders, predefined modes, and FDv2 polling/streaming data source builders, plus comprehensive tests.
> 
> - **Configuration**:
>   - Add internal `Configuration.DataSystem` and `ConfigurationBuilder.DataSystem(...)` wiring.
>   - Expose `Components.DataSystem()` (internal) returning `DataSystemModes`.
> - **Data system (internal)**:
>   - New builders and types: `DataSystemBuilder`, `DataSystemModes`, `DataSystemComponents`, `DataSystemConfiguration` (with `Initializers`, `Synchronizers`, `FDv1FallbackSynchronizer`, and persistent store with `ReadOnly/ReadWrite` modes).
>   - New FDv2 data source builders: `FDv2StreamingDataSourceBuilder` and `FDv2PollingDataSourceBuilder` with endpoint overrides and tuning options.
> - **Tests**:
>   - Extend `ConfigurationTest` with cases covering default/streaming/polling/daemon/persistent-store/custom data system configurations and list replacement/aggregation behaviors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8d48a680e1eea87611f50ebe7f713bb0b7e3a848. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


pkgs/sdk/server/src/Integrations/DataSystemComponents.cs
```diff
@@ -0,0 +1,47 @@
...
+namespace LaunchDarkly.Sdk.Server.Integrations
+{
+    /// <summary>
+    /// Components for use with the data system.
+    /// <para>
+    /// This class is not stable, and not subject to any backwards compatibility guarantees or semantic versioning.
+    /// It is not suitable for production usage. Do not use it. You have been warned.
+    /// </para>
+    /// </summary>
+    internal static class DataSystemComponents
```
kinyoklion: It did not feel reasonable to add this to the ever growing list of "Components".

pkgs/sdk/server/src/Integrations/DataSystemModes.cs
```diff
@@ -0,0 +1,159 @@
...
+    /// <summary>
+    /// A set of different data system modes which provided pre-configured <see cref="DataSystemBuilder"/>s.
+    /// <para>
+    /// This class is not stable, and not subject to any backwards compatibility guarantees or semantic versioning.
+    /// It is not suitable for production usage. Do not use it. You have been warned.
+    /// </para>
+    /// </summary>
+    internal sealed class DataSystemModes
+    {
+        // This implementation is non-static to allow for easy usage with "Components".
```
kinyoklion: It would be nice to have this static, but then it doesn't really fit the components pattern at all. This at least makes the data system discoverable at that level. It also isn't a performance concern as this isn't a high frequency operation.

pkgs/sdk/server/src/Integrations/DataSystemBuilder.cs
```diff
@@ -0,0 +1,112 @@
...
+            return this;
+        }
+
+        /// <summary>
+        /// Configured the FDv1 fallback synchronizer.
+        /// <remarks>LaunchDarkly can instruct the SDK to fall back to this synchronizer.</remarks>
+        /// </summary>
+        /// <param name="fdv1FallbackSynchronizer">the FDv1 fallback synchronizer</param>
+        /// <returns>a reference to the builder</returns>
+        public DataSystemBuilder FDv1FallbackSynchronizer(IComponentConfigurer<IDataSource> fdv1FallbackSynchronizer)
```
tanderson-ld: I'm not sure if this should be publicly exposed? It seems like a internal concern.

pkgs/sdk/server/src/Subsystems/DataSystemConfiguration.cs
```diff
@@ -0,0 +1,81 @@
...
+            /// <summary>
+            /// The data system can read from, and write to, the persistent store.
+            /// </summary>
+            ReadWrite,
+        }
+
+        /// <summary>
+        /// A list of factories for creating data sources for initialization.
+        /// </summary>
+        public IReadOnlyList<IComponentConfigurer<IDataSource>> Initializers { get; }
```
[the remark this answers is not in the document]
kinyoklion: Seems fine. We may have to capture some additional context via closure, factory implementation.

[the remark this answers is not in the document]
kinyoklion: Externally I think we use the IComponentConfigurer just because it is what these things currently work with. FWIW I think it is the best interface.

PR #194, opened 2025-12-10
chore: Add FDv1 DataSystem.

kinyoklion:
This is the first change which impacts the FDv1 data path. The code components are all the same as they were before, but the implementation has been moved inside an FDv1 data system.

Previously the client had direct access to each of the components involved, the data store, the data source, and the various updaters. This moves those all into the "DataSystem" and abstracts the clients access to them to only what is needed.

Some tests directly accessed these members as an internal, I would like to continue testing that functionality for now, but I didn't want as much exposure. So I have made a "Testing" member of the FDv1 data system for now. 

I think we should replace these tests with behavioral tests in the future.

---

> [!NOTE]
> Introduces FDv1 DataSystem to wrap data source/store and updates LdClient, flag tracking, and tests to use the new abstractions.
> 
> - **Core Data System**
>   - Add `FDv1DataSystem` implementing `IDataSystem` to own `IDataSource`, `IDataStore`, status providers, start/initialized, and disposal.
>   - Add `ReadonlyStoreFacade` (`IReadOnlyStore`) for read-only access to `IDataStore` (`Get`, `GetAll`, `Initialized`, `GetMetadata`).
>   - Add `FlagChangedFacade` implementing `IFlagChanged` proxying `FlagChanged` events from `DataSourceUpdatesImpl`.
> - **LdClient**
>   - Replace direct usage of `IDataSource`/`IDataStore` with `_dataSystem` (`FDv1DataSystem.Create`).
>   - Route `DataSourceStatusProvider`, `DataStoreStatusProvider`, `Initialized`, startup, and disposal via `_dataSystem`.
>   - Read flags/segments via `_dataSystem.Store` in evaluations and `AllFlagsState`.
>   - Construct `FlagTrackerImpl` with `_dataSystem.FlagChanged`.
> - **Flag Tracking**
>   - `FlagTrackerImpl` now depends on `IFlagChanged` (not `DataSourceUpdatesImpl`).
> - **Tests**
>   - Update tests to assert data source type via `FDv1DataSystem.Testing.DataSource` and to use `FlagChangedFacade` with `FlagTrackerImpl`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f627b22b6f4c23442bc43540a2d98f0483e91fa9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkgs/sdk/server/src/LdClient.cs
+++ b/pkgs/sdk/server/src/LdClient.cs
@@ -206,7 +206,7 @@ namespace LaunchDarkly.Sdk.Server
 
             var initTask = _dataSystem.Start();
 
-            if (_dataSystem.Initialized)
+            if (!_dataSystem.Initialized)
             {
                 _log.Info(InitWaitTimeInfo, _configuration.StartWaitTime.TotalMilliseconds);
                 if (_configuration.StartWaitTime >= ExcessiveInitWaitTime)


pkgs/sdk/server/src/Internal/DataSystem/IDataSystem.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        DataStoreTypes.ItemDescriptor? Get(DataStoreTypes.DataKind kind, string key);
+        DataStoreTypes.KeyedItems<DataStoreTypes.ItemDescriptor> GetAll(DataStoreTypes.DataKind kind);
+
+        bool Initialized();
+
+        DataStoreTypes.InitMetadata GetMetadata();
+    }
+
+    internal interface IFlagChanged
```
kinyoklion: This was only available on the concrete implementation of data source updates, so the flag tracker was dependent on the concrete implementation.

pkgs/sdk/server/src/Internal/DataSystem/IDataSystem.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        event EventHandler<FlagChangeEvent> FlagChanged;
+    }
+
+    internal interface IDataSystem
```
kinyoklion: This interface isn't ideal by the Law of Demeter, but it maintains compatibility with the external interface we currently have.

pkgs/sdk/server/src/LdClient.cs
kinyoklion: I need to undo formatting in this file. My format reflex is too strong.

PR #208, opened 2025-12-18
chore: updates datasources to report recoverability in errors

tanderson-ld:
**Requirements**

- [x] I have added test coverage for new or changed functionality
Didn't do yet, going to do as part of overall benchtesting before release.

Adds tracking of Recover-ability on data source ErrorInfo so that can be communicated outward by datasources.  Then the FDv2 datasource can blacklist conditionally on the recover-ability of the error.  This now allows data sources to report Off (intentional shutdown) without getting blacklisted.

---

> [!NOTE]
> Add `Recoverable` to `DataSourceStatus.ErrorInfo` and propagate it across all data sources to drive `Interrupted` vs `Off`, conditional FDv2 blacklisting/fallback, and improved shutdown handling; update tests accordingly.
> 
> - **Interfaces**
>   - Add `bool Recoverable` to `DataSourceStatus.ErrorInfo` and include it in `FromException`/`FromHttpError`.
> - **Data Sources**
>   - **Polling/Streaming (FDv1 & FDv2)**:
>     - Compute recoverability via `HttpErrors.IsRecoverable` and populate `ErrorInfo` (set `Recoverable=true` for `InvalidData`/`StoreError`).
>     - Route status: `Interrupted` for recoverable, `Off` for unrecoverable; pass `ErrorInfo` to status updates.
>     - Add `Shutdown(ErrorInfo?)` with atomic guard; update `Dispose` to call shutdown and avoid duplicate work; ensure `Off` status on shutdown.
>     - For unrecoverable init errors, complete init with `false`.
>   - **FDv2**:
>     - Honor `FDv1Fallback` header; blacklist current synchronizer only when error is unrecoverable.
> - **Tests**
>   - Update/extend tests to assert `Recoverable` semantics, `FDv1Fallback`, status transitions (`Interrupted` vs `Off`), init task results, and shutdown behavior; add waits for async status updates.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c56e4077d89a8129814769cf69bb43f2635b8b5d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkgs/sdk/server/test/LdClientEndToEndTest.cs
+++ b/pkgs/sdk/server/test/LdClientEndToEndTest.cs
@@ -57,6 +57,9 @@ namespace LaunchDarkly.Sdk.Server
                 using (var client = new LdClient(config))
                 {
                     Assert.False(client.Initialized);
+                    // Wait for the status to be updated to Off, as the status update happens asynchronously
+                    var statusUpdated = client.DataSourceStatusProvider.WaitFor(DataSourceState.Off, TimeSpan.FromSeconds(5));
+                    Assert.True(statusUpdated, "Status should have been updated to Off");
                     Assert.Equal(DataSourceState.Off, client.DataSourceStatusProvider.Status.State);
 
                     var value = client.BoolVariation(AlwaysTrueFlag.Key, BasicUser, false);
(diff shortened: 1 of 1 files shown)


pkgs/sdk/server/src/Internal/DataSources/PollingDataSource.cs
```diff
@@ -134,11 +136,17 @@ private void Dispose(bool disposing)
...
-                _featureRequestor.Dispose();
+                Shutdown(null);
             }
         }
 
+        private void Shutdown(DataSourceStatus.ErrorInfo? errorInfo)
+        {
+            _canceller?.Cancel();
+            _featureRequestor.Dispose();
+            _dataSourceUpdates.UpdateStatus(DataSourceState.Off, errorInfo);
```
kinyoklion: Emits events during dispose feels like something we shouldn't be doing. Because the thing we are emitting to could already be disposed/finalized.

kinyoklion (approved):
I would prefer no data source updates triggered by dispose. But everything else looks good.

PR #218, opened 2026-01-09
fix: Retry after partial file reads.

kinyoklion:
When we detect a file has been written there isn't any promise that the file content itself will actually be complete, or that a subsequent notification for that file will be received when it is. 

Currently sometimes the file update will trigger, and the JSON parsing will fail, and then it won't get re-triggered. Which causes some tests to fail.

This introduces a retry for any failed JSON parsing up to a maximum number of times.

Ideally we would extract the file watcher and abstract the file system API so we could simulate these conditions.

---

> [!NOTE]
> **Overview**
> When **auto-update** is enabled, `FileDataSource` no longer treats a failed parse (or transient read error during a retry) as final if the file may have been read mid-write. It retries up to **5 times** with a **600 ms** delay, then stops until the next file-change notification or another external load (`Start` / watcher).
> 
> Retry logic is **per failure episode**: each watcher-triggered or `Start`-driven load clears per-path failure counts. A single scheduled retry chain re-reads all paths; redundant retries are skipped if a successful load already cleared failure state. **Dispose** blocks further loads from pending retries. With **auto-update off**, behavior stays one-shot (no background retries).
> 
> Public docs on `FileDataSourceBuilder.AutoUpdate` are updated to describe give-up-after-repeated-failures instead of indefinite retry.
> 
> Tests add scripted file readers and cover retry success, max attempts, budget reset, alternate parsers, read errors during retry, dispose, multi-path abandonment, and looser reload assertions where watcher timing makes exact versions non-deterministic.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b9fcb6368e0e520895be494b9aa324bbf8d6a514. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkgs/sdk/server/src/Internal/DataSources/FileDataSource.cs
+++ b/pkgs/sdk/server/src/Internal/DataSources/FileDataSource.cs
@@ -28,6 +28,10 @@ namespace LaunchDarkly.Sdk.Server.Internal.DataSources
         private volatile int _lastVersion;
         private object _updateLock = new object();
 
+        private const int MaxRetries = 5;
+        private readonly TimeSpan RetryDelay = TimeSpan.FromSeconds(0.6);
+        private readonly Dictionary<string, int> _retryCounts = new Dictionary<string, int>();
+
         public FileDataSource(IDataSourceUpdates dataSourceUpdates, FileDataTypes.IFileReader fileReader,
             List<string> paths, bool autoUpdate, Func<string, object> alternateParser, bool skipMissingPaths,
             FileDataTypes.DuplicateKeysHandling duplicateKeysHandling,
@@ -102,10 +106,40 @@ namespace LaunchDarkly.Sdk.Server.Internal.DataSources
                         _logger.Debug("file data: {0}", content);
                         var data = _parser.Parse(content, version);
                         _dataMerger.AddToData(data, flags, segments);
+                        // Remove any retry count associated with this path.
+                        _retryCounts.Remove(path);
                     }
                     catch (FileNotFoundException) when (_skipMissingPaths)
                     {
                         _logger.Debug("{0}: {1}", path, "File not found");
+                    }
+                    catch (System.Text.Json.JsonException)
+                    {
+                        // We may have received the notification of a file change while the file was being written.
+                        // So we may read an empty or partially written file. So, when we encounter a JSON parsing issue
+                        // we will retry after a short delay.
+                        // We will retry up to MaxRetries times before giving up.
+                        if (!_retryCounts.ContainsKey(path))
+                        {
+                            _retryCounts[path] = 0;
+                        }
+                        _retryCounts[path]++;
+
+                        if (_retryCounts[path] < MaxRetries)
+                        {
+                            _logger.Warn("{0}: {1}", path, "Failed to parse file, retrying in " + RetryDelay.TotalMilliseconds + " milliseconds");
+                            Task.Run(async () =>
+                            {
+                                await Task.Delay(RetryDelay);
+                                LoadAll();
+                            });
+                        }
+                        else
+                        {
+                            _logger.Error("{0}: {1}", path, "Failed to parse file after " + MaxRetries + " retries");
+                        }
+
+                        return;
                     }
                     catch (Exception e)
                     {
@@ -113,6 +147,10 @@ namespace LaunchDarkly.Sdk.Server.Internal.DataSources
                         return;
                     }
                 }
+                
+                // If any files failed to load, from anything other than not existing, then that
+                // update would fail. This behavior is retained with the addition of the retry. But it should be
+                // examined.
 
                 var allData = new FullDataSet<ItemDescriptor>(
                     ImmutableDictionary.Create<DataKind, KeyedItems<ItemDescriptor>>()
--- a/pkgs/sdk/server/src/Internal/DataSources/FileDataSource.cs
+++ b/pkgs/sdk/server/src/Internal/DataSources/FileDataSource.cs
@@ -25,11 +25,13 @@ namespace LaunchDarkly.Sdk.Server.Internal.DataSources
         private readonly Logger _logger;
         private volatile bool _started;
         private volatile bool _loadedValidData;
+        private volatile bool _disposed;
         private volatile int _lastVersion;
         private object _updateLock = new object();
 
         private const int MaxRetries = 5;
-        private readonly TimeSpan RetryDelay = TimeSpan.FromSeconds(0.6);
+        private static readonly TimeSpan RetryDelay = TimeSpan.FromMilliseconds(600);
+        // Per-path JSON-parse retry counters. Only touched inside _updateLock.
         private readonly Dictionary<string, int> _retryCounts = new Dictionary<string, int>();
 
         public FileDataSource(IDataSourceUpdates dataSourceUpdates, FileDataTypes.IFileReader fileReader,
@@ -87,6 +89,7 @@ namespace LaunchDarkly.Sdk.Server.Internal.DataSources
         {
             if (disposing)
             {
+                _disposed = true;
                 _reloader?.Dispose();
             }
         }
@@ -95,6 +98,10 @@ namespace LaunchDarkly.Sdk.Server.Internal.DataSources
         {
             lock (_updateLock)
             {
+                if (_disposed)
+                {
+                    return;
+                }
                 var version = Interlocked.Increment(ref _lastVersion);
                 var flags = new Dictionary<string, ItemDescriptor>();
                 var segments = new Dictionary<string, ItemDescriptor>();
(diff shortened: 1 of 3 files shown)


pkgs/sdk/server/src/Internal/DataSources/FileDataSource.cs
```diff
@@ -83,14 +96,26 @@ private void Dispose(bool disposing)
         {
             if (disposing)
             {
+                _disposed = true;
```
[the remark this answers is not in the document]
kinyoklion: I am Claude, an AI assistant. I reply for the author.

We accept this risk. The window is small and bounded. A load that passed the check completes under the lock. At most one scheduled retry runs after Dispose. That retry sees _disposed under the lock and stops. DataSourceUpdatesImpl catches a late Init and logs it. The same window exists on main for watcher callbacks. A lock in Dispose can block for up to 30 seconds behind FlagFileReader. That cost is larger than the risk. No change is planned.

pkgs/sdk/server/src/Internal/DataSources/FileDataSource.cs
```diff
@@ -100,7 +125,21 @@ private void LoadAll()
...
+                        catch (Exception e)
+                        {
+                            // A file-change notification can fire while the file is mid-write, so a parse
+                            // failure may just mean we read an empty or partially written file. This applies
+                            // to any configured parser (JSON or alternate), so we treat every failure of
+                            // Parse — as opposed to reading the file — as potentially transient.
+                            HandleParseFailure(path, e);
+                            return;
+                        }
+                        _parseFailureCounts.Remove(path);
```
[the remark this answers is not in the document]
kinyoklion: I am Claude, an AI assistant. I reply for the author.

This is fixed in 64cb5142. A read failure on a retry now uses the same per-path budget. The chain schedules the next retry until the budget is used. Then the episode ends with terminal error logs. External loads keep the old behavior for non-parse errors. A new test covers this path. The test fails on the old code.

pkgs/sdk/server/src/Internal/DataSources/FileDataSource.cs
```diff
@@ -22,12 +22,24 @@ internal sealed class FileDataSource : IDataSource
...
+        private const int MaxParseAttempts = 5;
+        private static readonly TimeSpan RetryDelay = TimeSpan.FromMilliseconds(600);
+        // Consecutive parse failures per path within the current failure episode. An externally
+        // triggered load (Start or a file-change notification) starts a new episode and clears this,
+        // so the retry budget is per-episode, not per-lifetime. Only touched inside _updateLock.
+        private readonly Dictionary<string, int> _parseFailureCounts = new Dictionary<string, int>();
+        // Whether a delayed retry is already scheduled; at most one retry chain exists at a time,
+        // since each retry re-reads every path anyway. Only touched inside _updateLock.
+        private bool _retryPending;
+
```
tanderson-ld, line 43: Missing comment doc for constructor params.

pkgs/sdk/server/src/Internal/DataSources/FileDataSource.cs
```diff
@@ -22,12 +22,24 @@ internal sealed class FileDataSource : IDataSource
...
         private readonly bool _skipMissingPaths;
+        private readonly bool _autoUpdate;
         private readonly Logger _logger;
         private volatile bool _started;
         private volatile bool _loadedValidData;
+        private volatile bool _disposed;
         private volatile int _lastVersion;
         private object _updateLock = new object();
 
+        private const int MaxParseAttempts = 5;
```
tanderson-ld: Does this take the probability of the parsing incomplete data failure to 0 or just reduce it a bunch and nearly to 0?

pkgs/sdk/server/src/Internal/DataSources/FileDataSource.cs
```diff
@@ -124,12 +172,123 @@ private void LoadAll()
             }
         }
 
+        // Called under _updateLock when parsing a path's content fails. Since an externally
+        // triggered load clears _parseFailureCounts before reading, any existing count for the
+        // path belongs to the current episode.
+        private void HandleParseFailure(string path, Exception e)
+        {
+            if (!_autoUpdate)
```
tanderson-ld, line 210: I know it is probably more complex to handle, but should it be willing to retry even when auto update is off?  I guess this is the distinction between "load once (w failure)" vs "load once (success)".  Which is "loaded once" ?

In the customer's shoes, I'd rather it retry if there was a race condition with whatever was updating the file (maybe an external sync process) even if I had auto update off.

tanderson-ld (commented):
I'm still reviewing for lock safety and other scheduling races / order of operations type issues, just leaving this so others know it is getting review.

pkgs/sdk/server/src/Internal/DataSources/FileDataSource.cs
```diff
@@ -124,12 +172,123 @@ private void LoadAll()
...
+                }
+                catch (Exception e)
+                {
+                    // Nothing observes this task, so any escaping exception would otherwise vanish.
+                    LogHelpers.LogException(_logger, "Unexpected error while retrying file data load", e);
+                }
+            });
+        }
+
+        private void RetryLoadAll()
```
tanderson-ld: Having RetryLoadAll and LoadAll with a isRetry param seems redundant.  Consider getting rid of RetryLoadAll and moving logic into LoadAll and check when isRetry == true.  There is already the inverse check of `if (!isRetry)` in LoadAll, so this pending clearing stuff could live in the else branch.

Also the dispose check already exists in LoadAll.

pkgs/sdk/server/src/Internal/DataSources/FileDataSource.cs
```diff
@@ -124,12 +172,123 @@ private void LoadAll()
...
+                return;
+            }
+
+            var attempts = _parseFailureCounts.TryGetValue(path, out var previousAttempts)
+                ? previousAttempts + 1
+                : 1;
+
+            if (attempts < MaxParseAttempts)
+            {
+                _parseFailureCounts[path] = attempts;
```
tanderson-ld: I think the incrementing should be unconditional to be safer.

pkgs/sdk/server/src/Internal/DataSources/FileDataSource.cs
```diff
@@ -124,12 +172,123 @@ private void LoadAll()
...
+            {
+                // With auto-update off, files are documented to be loaded only once, so we don't
+                // retry in the background — Start()'s result stays final.
+                LogHelpers.LogException(_logger, "Failed to parse " + path, e);
+                return;
+            }
+
+            var attempts = _parseFailureCounts.TryGetValue(path, out var previousAttempts)
+                ? previousAttempts + 1
+                : 1;
```
tanderson-ld: Using 1 here when not found is less readable than defaulting to 0 and then unconditionally incrementing.  For a second I had to pause and wonder if 1 was actually the initial value, but it isn't, 0 is.

pkgs/sdk/server/src/Internal/DataSources/FileDataSource.cs
```diff
@@ -124,12 +172,123 @@ private void LoadAll()
...
+        }
+
+        // Called under _updateLock when a retry attempt fails before parsing (for example, a
+        // transient read error). The caller has already logged the exception; this charges the
+        // failure to the path's per-episode budget and continues or ends the retry chain.
+        private void HandleRetryLoadFailure(string path)
+        {
+            var attempts = _parseFailureCounts.TryGetValue(path, out var previousAttempts)
+                ? previousAttempts + 1
+                : 1;
```
tanderson-ld: Same comment here.

pkgs/sdk/server/src/Internal/DataSources/FileDataSource.cs
```diff
@@ -124,12 +172,123 @@ private void LoadAll()
...
+                    string.Format("{0}: Failed to parse file after {1} attempts", path, MaxParseAttempts), e);
+            }
+        }
+
+        // Called under _updateLock when a retry attempt fails before parsing (for example, a
+        // transient read error). The caller has already logged the exception; this charges the
+        // failure to the path's per-episode budget and continues or ends the retry chain.
+        private void HandleRetryLoadFailure(string path)
+        {
+            var attempts = _parseFailureCounts.TryGetValue(path, out var previousAttempts)
```
tanderson-ld: Is `_parseFailureCounts` too specific in that it calls out parsing?  This path can be hit for a general exception.

pkgs/sdk/server/src/Internal/DataSources/FileDataSource.cs
kinyoklion: There isn't really any way to make it 0. But for practical purposes it will effectively make it 0. It is an arbitrary number.

If you are writing a file, then probably the first retry would always fix it. But if you just continually stream JSON into a file, then there isn't any point at which it would be safely parse-able.

pkgs/sdk/server/src/Internal/DataSources/FileDataSource.cs
kinyoklion: I would rather not change that contract right now. Not that I know if anyone depends on it, but if someone had a problem, then I think we could add the option. Though there isn't really much of a reason to not just use autoUpdate in that case.

tanderson-ld (approved):

PR #249, opened 2026-04-15
feat!: Add per-execution runId, at-most-once tracking, and cross-process tracker resumption

jsonbailey:
BEGIN_COMMIT_OVERRIDE
feat!: Add per-execution runId on all AI track event payloads for billing isolation
feat!: Enforce at-most-once tracking — each metric type (duration, tokens, feedback, success/error, time-to-first-token) records once per tracker; duplicates are dropped with a warning
feat!: Add ResumptionToken on ILdAiConfigTracker and ILdAiClient.CreateTracker(resumptionToken, context) for cross-process tracker reconstruction with the original runId
feat: Add MetricSummary property on ILdAiConfigTracker summarizing recorded metrics
END_COMMIT_OVERRIDE

## Summary

- **Per-execution runId**: Every tracker gets a unique `runId` (UUID) included in all track event payloads, enabling per-execution billing isolation
- **At-most-once tracking**: Each metric type (duration, tokens, feedback, success/error, time-to-first-token) can only be tracked once per tracker instance — duplicates are dropped with a log warning
- **`CreateTracker()` on `LdAiConfig`**: Each call on a config (e.g. `LdAiCompletionConfig`) returns a new tracker with a fresh `runId` and independent tracking state. The returned tracker is always non-null, even when the config is disabled
- **`ResumptionToken` property on `ILdAiConfigTracker`**: URL-safe Base64-encoded JSON containing `{ runId, configKey, variationKey, version }` for cross-process tracker reconstruction
- **`CreateTracker(string token, Context context)` on `ILdAiClient`**: Reconstructs a tracker from a resumption token with the original `runId`, enabling deferred feedback from a different process. Model and provider names are set to empty strings on reconstruction
- **`MetricSummary` property on `ILdAiConfigTracker`**: Snapshot of metrics recorded on this tracker

## Test plan

- [x] `CreateTrackerReturnsTrackerWhenDisabled` — factory returns a non-null tracker for disabled configs
- [x] `CreateTrackerReturnsNewTrackerWithFreshRunId` — each tracker gets a unique runId
- [x] `CreateTrackerReturnsTrackerWithIndependentTrackingState` — at-most-once state is independent per tracker
- [x] `CreateTrackerCanBeCalledMultipleTimes` — factory produces multiple independent trackers
- [x] `ResumptionTokenContainsExpectedFields` — token contains exactly runId, configKey, variationKey, version
- [x] `ResumptionTokenIsUrlSafeBase64` — no `+`, `/`, or `=` characters
- [x] `ResumptionTokenIsConsistentAcrossCalls` — same token on repeated access
- [x] `CreateTrackerFromResumptionTokenRoundTrips` — full round-trip preserves runId
- [x] `CreateTrackerFromResumptionTokenSetsEmptyModelAndProvider` — reconstructed tracker has empty model/provider
- [x] `CreateTrackerFromInvalidTokenThrows` — malformed input throws ArgumentException
- [x] `CreateTrackerFromNullTokenThrows` — null input throws ArgumentNullException
- [x] `CreateTrackerFromTokenMissingRunIdThrows` — incomplete payload throws ArgumentException

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes public tracking semantics (runId on events, duplicate metrics ignored) and default version behavior, which can affect downstream analytics and billing correlation; resumption token parsing is a new trust boundary for cross-process flows.
> 
> **Overview**
> **AI config trackers are now scoped to a single run** with a UUID **`runId`** on every track payload, **`ResumptionToken`** / **`ILdAiClient.CreateTracker(token, context)`** for cross-process reuse of that run, and a **`MetricSummary`** snapshot of what was recorded.
> 
> **`LdAiConfigTracker`** is rebuilt around explicit fields (not the whole config): each **`CreateTracker()`** gets a new **`runId`**; duplicate **`Track*`** calls are dropped (with warnings) using atomic slots, including shared success/error; empty token usage does not consume the tokens slot. Track metadata adds **`runId`**, omits empty **`variationKey`**, and resumption tokens are URL-safe Base64 JSON without model/provider.
> 
> **Smaller contract tweaks:** caller defaults use **`version: 1`**; missing **`_ldMeta.version`** defaults to **1**; **`ModelConfig`** stores parameters in **`ImmutableDictionary`** so downcasts cannot mutate maps.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f7488be4987819c417e83b9ffa80065b35c360b8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkgs/sdk/server-ai/src/Config/LdAiConfig.cs
+++ b/pkgs/sdk/server-ai/src/Config/LdAiConfig.cs
@@ -1,6 +1,8 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using LaunchDarkly.Sdk.Server.Ai.DataModel;
+using LaunchDarkly.Sdk.Server.Ai.Interfaces;
 
 namespace LaunchDarkly.Sdk.Server.Ai.Config;
 
@@ -217,7 +219,8 @@ public record LdAiConfig
     /// </summary>
     public readonly ModelProvider Provider;
 
-    internal LdAiConfig(bool enabled, IEnumerable<Message> messages, Meta meta, Model model, Provider provider)
+    internal LdAiConfig(bool enabled, IEnumerable<Message> messages, Meta meta, Model model, Provider provider,
+        Func<ILdAiConfigTracker> createTracker = null)
     {
         Model = new ModelConfiguration(model?.Name ?? "", model?.Parameters ?? new Dictionary<string, LdValue>(),
             model?.Custom ?? new Dictionary<string, LdValue>());
@@ -226,6 +229,18 @@ public record LdAiConfig
         Version = meta?.Version ?? 1;
         Enabled = enabled;
         Provider = new ModelProvider(provider?.Name ?? "");
+        CreateTracker = createTracker;
+    }
+
+    internal LdAiConfig(LdAiConfig source, Func<ILdAiConfigTracker> createTracker)
+    {
+        Model = source.Model;
+        Messages = source.Messages;
+        VariationKey = source.VariationKey;
+        Version = source.Version;
+        Enabled = source.Enabled;
+        Provider = source.Provider;
+        CreateTracker = createTracker;
     }
     internal LdValue ToLdValue()
     {
@@ -277,6 +292,14 @@ public record LdAiConfig
     /// This field meant for internal LaunchDarkly usage.
     /// </summary>
     public int Version { get; }
+
+    /// <summary>
+    /// A factory that creates a new <see cref="ILdAiConfigTracker"/> with a fresh runId.
+    /// Each invocation returns a tracker with independent at-most-once tracking state.
+    /// This property is set when the config is returned by <see cref="LdAiClient.CompletionConfig"/>.
+    /// It will be null for configs created directly via the builder.
+    /// </summary>
+    public Func<ILdAiConfigTracker> CreateTracker { get; }
 
     /// <summary>
     /// Convenient helper that returns a disabled LdAiConfig.
--- a/pkgs/sdk/server-ai/src/Interfaces/ILdAiClient.cs
+++ b/pkgs/sdk/server-ai/src/Interfaces/ILdAiClient.cs
@@ -12,8 +12,9 @@ public interface ILdAiClient
 
     /// <summary>
     /// Retrieves a LaunchDarkly AI Completion Config identified by the given key. The return value
-    /// is an <see cref="ILdAiConfigTracker"/>, which makes the configuration available and
-    /// provides convenience methods for generating events related to model usage.
+    /// is an <see cref="LdAiConfig"/>, which makes the configuration available and provides
+    /// a <see cref="LdAiConfig.CreateTracker"/> factory for generating trackers capable of
+    /// emitting events related to model usage.
     ///
     /// Any variables provided will be interpolated into the prompt's messages.
     /// Additionally, the current LaunchDarkly context will be available as 'ldctx' within
@@ -25,8 +26,8 @@ public interface ILdAiClient
     /// <param name="defaultValue">the default config, if unable to retrieve from LaunchDarkly. When not provided,
     /// a disabled config is used as the fallback.</param>
     /// <param name="variables">the list of variables used when interpolating the prompt</param>
-    /// <returns>an AI Completion Config tracker</returns>
-    public ILdAiConfigTracker CompletionConfig(string key, Context context, LdAiConfig defaultValue = null,
+    /// <returns>an AI Completion Config</returns>
+    public LdAiConfig CompletionConfig(string key, Context context, LdAiConfig defaultValue = null,
         IReadOnlyDictionary<string, object> variables = null);
 
     /// <summary>
@@ -36,9 +37,9 @@ public interface ILdAiClient
     /// <param name="context">the context</param>
     /// <param name="defaultValue">the default config, if unable to retrieve from LaunchDarkly</param>
     /// <param name="variables">the list of variables used when interpolating the prompt</param>
-    /// <returns>an AI Completion Config tracker</returns>
+    /// <returns>an AI Completion Config</returns>
     [Obsolete("Use CompletionConfig instead.")]
-    public ILdAiConfigTracker Config(string key, Context context, LdAiConfig defaultValue = null,
+    public LdAiConfig Config(string key, Context context, LdAiConfig defaultValue = null,
         IReadOnlyDictionary<string, object> variables = null);
 
     /// <summary>
--- a/pkgs/sdk/server-ai/src/Interfaces/ILdAiClient.cs
+++ b/pkgs/sdk/server-ai/src/Interfaces/ILdAiClient.cs
@@ -40,4 +40,17 @@ public interface ILdAiClient
     [Obsolete("Use CompletionConfig instead.")]
     public ILdAiConfigTracker Config(string key, Context context, LdAiConfig defaultValue = null,
         IReadOnlyDictionary<string, object> variables = null);
+
+    /// <summary>
+    /// Reconstructs a tracker from a resumption token. This enables cross-process scenarios
+    /// such as deferred feedback, where a tracker's runId needs to be reused in a different
+    /// process or at a later time.
+    ///
+    /// The reconstructed tracker will have empty model and provider names, as these are not
+    /// included in the resumption token.
+    /// </summary>
+    /// <param name="resumptionToken">the resumption token obtained from <see cref="ILdAiConfigTracker.ResumptionToken"/></param>
+    /// <param name="context">the context to use for track events</param>
+    /// <returns>a tracker associated with the original runId</returns>
+    public ILdAiConfigTracker CreateTracker(string resumptionToken, Context context);
 }
--- a/pkgs/sdk/server-ai/src/Interfaces/ILdAiClient.cs
+++ b/pkgs/sdk/server-ai/src/Interfaces/ILdAiClient.cs
@@ -41,4 +41,17 @@ public interface ILdAiClient
     [Obsolete("Use CompletionConfig instead.")]
     public LdAiCompletionConfig Config(string key, Context context, LdAiCompletionConfigDefault defaultValue = null,
         IReadOnlyDictionary<string, object> variables = null);
+
+    /// <summary>
+    /// Reconstructs a tracker from a resumption token. This enables cross-process scenarios
+    /// such as deferred feedback, where a tracker's runId needs to be reused in a different
+    /// process or at a later time.
+    ///
+    /// The reconstructed tracker will have empty model and provider names, as these are not
+    /// included in the resumption token.
+    /// </summary>
+    /// <param name="resumptionToken">the resumption token obtained from <see cref="ILdAiConfigTracker.ResumptionToken"/></param>
+    /// <param name="context">the context to use for track events</param>
+    /// <returns>a tracker associated with the original runId</returns>
+    public ILdAiConfigTracker CreateTracker(string resumptionToken, Context context);
 }
--- a/pkgs/sdk/server-ai/src/Interfaces/ILdAiConfigTracker.cs
+++ b/pkgs/sdk/server-ai/src/Interfaces/ILdAiConfigTracker.cs
@@ -114,9 +114,7 @@ public interface ILdAiConfigTracker
     /// Creates a new tracker with a fresh runId for tracking a new execution.
     /// Each call returns a tracker with independent at-most-once tracking state,
     /// allowing multiple executions to be tracked separately against the same config.
-    ///
-    /// Returns null if the config is disabled.
     /// </summary>
-    /// <returns>a new tracker, or null if the config is disabled</returns>
+    /// <returns>a new tracker</returns>
     public ILdAiConfigTracker CreateTracker();
 }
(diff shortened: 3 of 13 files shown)


pkgs/sdk/server-ai/src/Interfaces/ILdAiClient.cs
```diff
@@ -36,8 +37,21 @@ public ILdAiConfigTracker CompletionConfig(string key, Context context, LdAiConf
...
+    public LdAiConfig Config(string key, Context context, LdAiConfig defaultValue = null,
         IReadOnlyDictionary<string, object> variables = null);
+
+    /// <summary>
+    /// Reconstructs a tracker from a resumption token. This enables cross-process scenarios
+    /// such as deferred feedback, where a tracker's runId needs to be reused in a different
+    /// process or at a later time.
+    ///
+    /// The reconstructed tracker will have empty model and provider names, as these are not
+    /// included in the resumption token.
```
kinyoklion, line 51: Is this a durable decision? Probably?

pkgs/sdk/server-ai/src/Interfaces/ILdAiConfigTracker.cs
```diff
@@ -6,21 +6,36 @@
 namespace LaunchDarkly.Sdk.Server.Ai.Interfaces;
 
 /// <summary>
-/// A utility capable of generating events related to a specific AI model
-/// configuration.
+/// Records metrics for a single AI run.
 /// </summary>
+/// <remarks>
+/// All events a tracker emits share a runId (a UUIDv4) so LaunchDarkly can correlate
```
kinyoklion:
```suggestion
/// All events emitted by a tracker share a runId (a UUIDv4) so LaunchDarkly can correlate
```

Maybe something like this. I had a hard time parsing the sentence the first time I read through it.

pkgs/sdk/server-ai/src/Interfaces/ILdAiConfigTracker.cs
```diff
@@ -6,21 +6,36 @@
...
+/// them in metrics views. See individual track methods for their specific semantics.
```
kinyoklion:
```suggestion
/// them. See individual track methods for their specific semantics.
```

Metric views may be one use-case, but what matters is we can correlate them.

pkgs/sdk/server-ai/src/LdAiConfigTracker.cs
```diff
@@ -91,17 +179,30 @@ public void TrackFeedback(Feedback feedback)
             default:
                 throw new ArgumentOutOfRangeException(nameof(feedback), feedback, null);
         }
+        _feedback = feedback;
     }
 
     /// <inheritdoc/>
     public void TrackSuccess()
     {
+        if (_trackedSuccess.HasValue)
```
kinyoklion: this would need to use an interlocked exchange to be thread-safe. (Or a lock, but an interlocked exchange would be better.)

kinyoklion (changes requested):

pkgs/sdk/server-ai/src/LdAiConfigTracker.cs
kinyoklion: Confirmed empirically with two xUnit tests using `Barrier`-aligned threads on `net8.0`. Both tests **fail on this branch** (race observed), and would pass once the slot fields are guarded with `Interlocked.CompareExchange`.

Results across 500 iterations each, 2 threads per iteration:
- `TrackSuccessIsAtMostOnceUnderConcurrency` — observed `$ld:ai:generation:success` emitted **twice** on iteration 13.
- `TrackSuccessAndTrackErrorAreMutuallyExclusiveUnderConcurrency` — **15 of 500** iterations emitted both a success AND an error event for the same `runId` (directly contradicts the interface docstring "TrackSuccess and TrackError share state; only one of the two can record per Tracker").

Tests, drop into `pkgs/sdk/server-ai/test/LdAiConfigTrackerTest.cs` (also need `using System.Linq;` and `using System.Threading;`):

```csharp
[Fact]
public void TrackSuccessIsAtMostOnceUnderConcurrency()
{
    const int iterations = 500;
    var maxObservedInvocations = 0;
    var iterationWithMax = -1;

    for (var i = 0; i < iterations; i++)
    {
        var mockClient = new Mock<ILaunchDarklyClient>();
        var context = Context.New("key");
        var tracker = MakeTracker(mockClient, "key", context);

        var barrier = new Barrier(2);
        var t1 = Task.Run(() => { barrier.SignalAndWait(); tracker.TrackSuccess(); });
        var t2 = Task.Run(() => { barrier.SignalAndWait(); tracker.TrackSuccess(); });
        Task.WaitAll(t1, t2);

        var calls = mockClient.Invocations.Count(inv =>
            inv.Method.Name == "Track" &&
            (string)inv.Arguments[0] == "$ld:ai:generation:success");

        if (calls > maxObservedInvocations) { maxObservedInvocations = calls; iterationWithMax = i; }
    }

    Assert.True(maxObservedInvocations == 1,
        $"Expected at-most-once emission of $ld:ai:generation:success across {iterations} " +
        $"iterations, but observed {maxObservedInvocations} emissions on iteration {iterationWithMax}. " +
        "TrackSuccess's check-then-set on _trackedSuccess is not atomic.");
}

[Fact]
public void TrackSuccessAndTrackErrorAreMutuallyExclusiveUnderConcurrency()
{
    const int iterations = 500;
    var iterationsWithBoth = 0;

    for (var i = 0; i < iterations; i++)
    {
        var mockClient = new Mock<ILaunchDarklyClient>();
        var context = Context.New("key");
        var tracker = MakeTracker(mockClient, "key", context);

        var barrier = new Barrier(2);
        var t1 = Task.Run(() => { barrier.SignalAndWait(); tracker.TrackSuccess(); });
        var t2 = Task.Run(() => { barrier.SignalAndWait(); tracker.TrackError(); });
        Task.WaitAll(t1, t2);

        var successCalls = mockClient.Invocations.Count(inv =>
            inv.Method.Name == "Track" && (string)inv.Arguments[0] == "$ld:ai:generation:success");
        var errorCalls = mockClient.Invocations.Count(inv =>
            inv.Method.Name == "Track" && (string)inv.Arguments[0] == "$ld:ai:generation:error");

        if (successCalls > 0 && errorCalls > 0) iterationsWithBoth++;
    }

    Assert.True(iterationsWithBoth == 0,
        $"Expected TrackSuccess and TrackError to be mutually exclusive across {iterations} " +
        $"iterations, but {iterationsWithBoth} iterations emitted both success AND error " +
        "events for the same runId. The shared _trackedSuccess slot is not atomically guarded.");
}
```

The same race exists in `TrackDuration`, `TrackTimeToFirstToken`, `TrackFeedback`, and `TrackTokens` — each is a check-then-set on an unsynchronized nullable field.

pkgs/sdk/server-ai/src/Interfaces/ILdAiClient.cs
jsonbailey: We tried to cut down as much data as possible in the resumption token to be just the necessary for proper reporting since theses tokens could be sent client side.

pkgs/sdk/server-ai/src/Config/LdAiConfig.cs
```diff
@@ -278,6 +296,17 @@ internal LdValue ToLdValue()
...
+    /// <summary>
+    /// Returns a new tracker for a fresh AI run. Each call mints a new runId (a UUIDv4)
+    /// that LaunchDarkly uses to correlate the run's events in metrics views. Call this
+    /// once per AI run; metrics from different runIds cannot be combined.
+    /// </summary>
+    /// <remarks>
+    /// Returns null for configs created directly via the builder; only configs returned by
+    /// <see cref="LdAiClient.CompletionConfig"/> have a factory wired in.
+    /// </remarks>
+    public ILdAiConfigTracker CreateTracker() => _trackerFactory?.Invoke();
```
kinyoklion: Because we don't compile on a version of .Net with nullable reference types I don't really like having the option to return null.

pkgs/sdk/server-ai/src/LdAiConfigTracker.cs
```diff
@@ -29,36 +52,88 @@ public class LdAiConfigTracker : ILdAiConfigTracker
...
     /// <param name="context">the context</param>
-    /// <exception cref="ArgumentNullException"></exception>
-    public LdAiConfigTracker(ILaunchDarklyClient client, string configKey, LdAiConfig config, Context context)
+    /// <param name="modelName">the model name</param>
+    /// <param name="providerName">the provider name</param>
+    public LdAiConfigTracker(ILaunchDarklyClient client, string runId, string configKey,
+        string variationKey, int version, Context context, string modelName, string providerName)
     {
-        Config = config ?? throw new ArgumentNullException(nameof(config));
-        _client = client ?? throw new ArgumentNullException(nameof(client));
```
kinyoklion, line 42: Feels like the client still should be null checked,

pkgs/sdk/server-ai/src/LdAiConfigTracker.cs
```diff
@@ -137,17 +238,92 @@ public async Task<Response> TrackRequest(Task<Response> request)
...
+            case 3: base64 += "="; break;
+        }
+
+        ResumptionPayload payload;
+        try
+        {
+            var json = Encoding.UTF8.GetString(Convert.FromBase64String(base64));
+            payload = JsonSerializer.Deserialize<ResumptionPayload>(json);
+        }
+        catch (Exception e) when (e is FormatException || e is JsonException)
```
kinyoklion: Is there a reason we narrow here instead of any exception?

pkgs/sdk/server-ai/src/LdAiConfigTracker.cs
```diff
@@ -137,17 +238,92 @@ public async Task<Response> TrackRequest(Task<Response> request)
...
+        {
+            case 2: base64 += "=="; break;
+            case 3: base64 += "="; break;
+        }
+
+        ResumptionPayload payload;
+        try
+        {
+            var json = Encoding.UTF8.GetString(Convert.FromBase64String(base64));
+            payload = JsonSerializer.Deserialize<ResumptionPayload>(json);
```
kinyoklion, line 307: Does it matter than this will deserialize a missing version to 0, where the builder would use 1? Probably not as the 1 was just for configs before the version was added?

pkgs/sdk/server-ai/src/Config/LdAiConfig.cs
```diff
@@ -1,13 +1,15 @@
...
 using LaunchDarkly.Sdk.Server.Ai.DataModel;
+using LaunchDarkly.Sdk.Server.Ai.Interfaces;
 
 namespace LaunchDarkly.Sdk.Server.Ai.Config;
 
 /// <summary>
 /// Represents an AI Config, which contains model parameters and prompt messages.
 /// </summary>
-public record LdAiConfig
+public class LdAiConfig
```
kinyoklion: This would have had auto-generated equals, hashcode, tostring before. So we should consider if we need any of them. (It would have also worked with `with` expressions, but nothing we can do about that.)

pkgs/sdk/server-ai/src/LdAiConfigTracker.cs
```diff
@@ -29,36 +52,88 @@ public class LdAiConfigTracker : ILdAiConfigTracker
...
+        {
+            trackDataBuilder.Add("variationKey", LdValue.Of(_variationKey));
+        }
+        _trackData = LdValue.ObjectFrom(trackDataBuilder);
     }
 
+    /// <inheritdoc/>
+    public string ResumptionToken
+    {
+        get
```
kinyoklion: I am hesitant to have this be a property. Usually the expectation of a property is that it doesn't really do work, but this does comparatively substantial work?

kinyoklion:
Tangentially — not introduced by this PR, but surfaced while reviewing it — `ModelConfiguration.Parameters` and `.Custom` are typed as `IReadOnlyDictionary<>` but the underlying instances are mutable `Dictionary<>`s that the builder hands off by reference. A downstream consumer can downcast through any `IDictionary<>`-compatible interface and mutate them, and writes are visible on `config.Model.Parameters`.

Failing test that pins the desired behavior (it goes green once the dictionaries are stored as a defensive copy):

```csharp
// ModelConfiguration.Parameters is typed as IReadOnlyDictionary<>, so the
// public contract is "read-only". Even if a consumer downcasts to a
// mutable view, writes through that view must not be visible on
// config.Model.Parameters.
[Fact]
public void DowncastedMutationOfModelParametersMustNotAffectBuiltConfig()
{
    var config = LdAiConfig.New().SetModelParam("temperature", LdValue.Of(0.7)).Build();

    if (config.Model.Parameters is System.Collections.Generic.IDictionary<string, LdValue> mutable)
    {
        mutable["temperature"] = LdValue.Of(2.0);
    }

    Assert.Equal(0.7, config.Model.Parameters["temperature"].AsDouble);
}
```

Two options for the fix:

1. **`ImmutableDictionary<string, LdValue>`** in `ModelConfiguration`'s ctor (and copy the input into it). Type system enforces no mutation.
2. **Store the whole thing as `LdValue`** instead — `LdValue.ObjectFrom(...)` already handles immutability and deep-copies primitive map data.

Happy to file as a separate issue if it's out of scope for this PR.

kinyoklion (approved):

PR #251, opened 2026-04-30
fix: Honor x-ld-fd-fallback header in FDv2 initializer phase and on successful responses

keelerm84:
## Summary

- `FDv2PollingDataSource` and `FDv2StreamingDataSource` now honor `X-LD-FD-Fallback: true` on **successful** responses (not just errors): the payload is applied first, then the data source closes with `FDv1Fallback=true` so the action applier can swap to FDv1.
- The FDv1 fallback action applier is now wired to the **initializers** composite as well as the synchronizers composite, so a directive arriving during initialization bypasses remaining initializers and the FDv2 synchronizer chain. A new `FDv1FallbackLatch` + `GatedObserver` coordination ensures the existing fallback-on-Off appliers stop reacting after fallback is triggered, preventing them from blocking the now-running FDv1 fallback entry on stale exhaustion signals.
- Contract test service declares the `fdv1-fallback` capability and accepts a top-level `dataSystem.fdv1Fallback` config object, wiring it directly to `FDv1FallbackSynchronizer` instead of inferring from the last polling synchronizer. Bumps the FDv2 contract-tests pin from `v3.0.0-alpha.3` to `v3.0.0-alpha.6`.

Mirrors the Go reference change in [go-server-sdk#365](https://github.com/launchdarkly/go-server-sdk/pull/365) and contract-test wiring in [go-server-sdk#368](https://github.com/launchdarkly/go-server-sdk/pull/368). Spec rationale lives in [sdk-specs#155](https://github.com/launchdarkly/sdk-specs/pull/155); the new harness suite is in [sdk-test-harness#336](https://github.com/launchdarkly/sdk-test-harness/pull/336).

## Test plan

- [x] `dotnet test pkgs/sdk/server/test/...` — 1555 tests pass (8 new tests added). FDv2-only filter passes 159 across 3 consecutive runs.
- [x] New unit-level coverage for: polling/streaming success-with-header → payload-then-fallback, applier `BlockCurrent → DisposeCurrent → (GoToNext)*N → StartCurrent` sequencing, and `GatedObserver` suppression after the latch fires.
- [ ] Contract-test workflow against `v3.0.0-alpha.6` runs the new "FDv1 Fallback Directive" suite cleanly in CI.

### Test coverage caveat

There is **no in-repo end-to-end test** that drives a real `FDv2DataSource` composite through the initializer-phase fallback path (initializer reports `Off+FDv1Fallback` → payload applied → FDv2 synchronizer chain bypassed → FDv1 fallback engaged). An attempt at one was flaky against the existing CompositeSource action queue; rather than commit a flaky test, the unit-level applier and latch tests were kept and the contract-test harness's `fdv1-fallback` suite is the end-to-end safety net. The synchronizer-phase end-to-end fallback IS covered by the pre-existing `ErrorWithFDv1FallbackTriggersFallbackToFDv1Synchronizers` test.

---

> [!NOTE]
> **Medium Risk**
> Medium risk: changes core FDv2 data-source/composite switching and fallback behavior based on `x-ld-fd-fallback`, which could impact initialization and update delivery if edge cases are missed.
> 
> **Overview**
> **FDv2 now honors the `x-ld-fd-fallback: true` directive across more paths and cleanly hands off to FDv1.** Polling and streaming sources propagate the directive even on *successful* responses by tagging the emitted `ChangeSet` and then shutting down with `FDv1Fallback=true` so the composite can switch to the FDv1 fallback synchronizer.
> 
> The FDv2 composite fallback logic is reworked to be phase-agnostic: a new `CompositeEntryKind` plus `ICompositeSourceActionable.BlockAll` lets the FDv1 fallback applier remove all FDv2 entries in one shot (initializers and synchronizers) and advance to the FDv1 fallback entry, while other appliers bail out when the directive is present to avoid double-advancing.
> 
> Contract-test plumbing is updated to support the new harness suite: the test harness pin moves to `v3.0.0-alpha.6`, the test service advertises `fdv1-fallback` and accepts `dataSystem.fdv1Fallback` (with correct JSON name), and CI uploads the test-service log on FDv2 contract-test failures. Extensive new unit/integration tests cover directive-on-success, directive-on-error, initializer-phase skipping, and “no fallback configured” halt behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 752da6a509808322c280d15447a47ff104e669bb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkgs/sdk/server/src/Internal/DataSources/CompositeDataSource/CompositeSource.cs
+++ b/pkgs/sdk/server/src/Internal/DataSources/CompositeDataSource/CompositeSource.cs
@@ -451,6 +451,16 @@ namespace LaunchDarkly.Sdk.Server.Internal.DataSources
             });
         }
 
+        // Explicit interface implementation forwards to the private EnqueueAction so callers using the
+        // ICompositeSourceActionable abstraction can defer arbitrary work onto the same serialized
+        // action queue. This allows observers to interleave their advancement decisions with other
+        // queued actions, while keeping the underlying queue plumbing private.
+        void ICompositeSourceActionable.EnqueueAction(Action action)
+        {
+            if (action == null) return;
+            EnqueueAction(action);
+        }
+
         private void logTransition(String previousDescription, String currentDescription) {
             if (previousDescription != null && currentDescription != null) {
                 _log.Debug("{0} transitioned from {1} to {2}.", _compositeDescription, previousDescription, currentDescription);
--- a/pkgs/sdk/server/src/Internal/DataSources/CompositeDataSource/CompositeSource.cs
+++ b/pkgs/sdk/server/src/Internal/DataSources/CompositeDataSource/CompositeSource.cs
@@ -40,29 +40,17 @@ namespace LaunchDarkly.Sdk.Server.Internal.DataSources
         private IDataSource _currentDataSource;
 
         /// <summary>
-        /// Creates a new <see cref="CompositeSource"/>. Every entry is treated as kind
-        /// <see cref="CompositeEntryKind.FDv2"/>.
+        /// Creates a new <see cref="CompositeSource"/>. Each entry carries an explicit
+        /// <see cref="CompositeEntryKind"/> so appliers can express "block every FDv2 entry"
+        /// via <see cref="ICompositeSourceActionable.BlockAll"/> without needing to know list
+        /// positions. For inner sub-composites whose entries never participate in cross-kind
+        /// blocking, callers should pass <see cref="CompositeEntryKind.FDv2"/> uniformly.
         /// </summary>
         /// <param name="compositeDescription">description of the composite source for logging purposes</param>
         /// <param name="updatesSink">the sink that receives updates from the active source</param>
-        /// <param name="factoryTuples">the ordered list of source factories and their associated action applier factories</param>
+        /// <param name="factoryTuples">the ordered list of source factories, action applier factories, and entry kinds</param>
         /// <param name="logger">the logger instance to use</param>
         /// <param name="circular">whether to loop off the end of the list back to the start when fallback occurs</param>
-        public CompositeSource(
-            string compositeDescription,
-            IDataSourceUpdatesV2 updatesSink,
-            IList<(SourceFactory Factory, ActionApplierFactory ActionApplierFactory)> factoryTuples,
-            Logger logger,
-            bool circular = true)
-            : this(compositeDescription, updatesSink, WithDefaultKind(factoryTuples), logger, circular)
-        {
-        }
-
-        /// <summary>
-        /// Creates a new <see cref="CompositeSource"/> with explicit per-entry kinds. The kind is
-        /// surfaced to <see cref="ICompositeSourceActionable.BlockAll"/> so appliers can express
-        /// "block every FDv2 entry" without needing to know list positions.
-        /// </summary>
         public CompositeSource(
             string compositeDescription,
             IDataSourceUpdatesV2 updatesSink,
@@ -92,18 +80,6 @@ namespace LaunchDarkly.Sdk.Server.Internal.DataSources
                 circular: circular,
                 initialList: factoryTuples
             );
-        }
-
-        private static IList<(SourceFactory Factory, ActionApplierFactory ActionApplierFactory, CompositeEntryKind Kind)>
-            WithDefaultKind(IList<(SourceFactory Factory, ActionApplierFactory ActionApplierFactory)> tuples)
-        {
-            if (tuples is null) return null;
-            var result = new List<(SourceFactory, ActionApplierFactory, CompositeEntryKind)>(tuples.Count);
-            foreach (var t in tuples)
-            {
-                result.Add((t.Factory, t.ActionApplierFactory, CompositeEntryKind.FDv2));
-            }
-            return result;
         }
 
         /// <summary>
--- a/pkgs/sdk/server/src/Internal/DataSources/CompositeDataSource/CompositeSource.cs
+++ b/pkgs/sdk/server/src/Internal/DataSources/CompositeDataSource/CompositeSource.cs
@@ -30,17 +30,18 @@ namespace LaunchDarkly.Sdk.Server.Internal.DataSources
 
         private readonly IDataSourceUpdatesV2 _originalUpdateSink;
         private readonly IDataSourceUpdatesV2 _sanitizedUpdateSink;
-        private readonly SourcesList<(SourceFactory Factory, ActionApplierFactory ActionApplierFactory)> _sourcesList;
+        private readonly SourcesList<(SourceFactory Factory, ActionApplierFactory ActionApplierFactory, CompositeEntryKind Kind)> _sourcesList;
         private readonly DisableableDataSourceUpdatesTracker _disableableTracker;
 
         // Tracks the entry from the sources list that was used to create the current
         // data source instance. This allows operations such as blacklist to remove
         // the correct factory/action-applier-factory tuple from the list.
-        private (SourceFactory Factory, ActionApplierFactory ActionApplierFactory) _currentEntry;
+        private (SourceFactory Factory, ActionApplierFactory ActionApplierFactory, CompositeEntryKind Kind) _currentEntry;
         private IDataSource _currentDataSource;
 
         /// <summary>
-        /// Creates a new <see cref="CompositeSource"/>.
+        /// Creates a new <see cref="CompositeSource"/>. Every entry is treated as kind
+        /// <see cref="CompositeEntryKind.FDv2"/>.
         /// </summary>
         /// <param name="compositeDescription">description of the composite source for logging purposes</param>
         /// <param name="updatesSink">the sink that receives updates from the active source</param>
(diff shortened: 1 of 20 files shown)


pkgs/sdk/server/src/Internal/FDv2DataSources/FDv2DataSource.cs
```diff
@@ -35,13 +35,28 @@ public static IDataSource CreateFDv2DataSource(
...
-            ActionApplierFactory fdv1FallbackApplierFactory = (actionable) => new FDv1FallbackActionApplier(actionable);
+                (actionable) => new GatedObserver(
+                    new ActionApplierTimedFallbackAndRecovery(actionable, sublogger), fdv1FallbackTriggered);
+
+            // From the synchronizers entry, the FDv1 fallback entry is the next entry in the
+            // outer list, so no extra entries to skip.
+            ActionApplierFactory fdv1FallbackApplierFactory =
+                (actionable) => new FDv1FallbackActionApplier(actionable, fdv1FallbackTriggered);
+            // From the initializers entry, the FDv1 fallback entry is two ahead when synchronizers
+            // are configured (skip past synchronizers), or one ahead when they are not.
```
tanderson-ld: This seems fragile.

pkgs/sdk/server/src/Internal/DataSources/CompositeDataSource/ICompositeSourceActionable.cs
```diff
@@ -34,10 +35,22 @@ internal interface ICompositeSourceActionable
...
+        /// Enqueues an arbitrary action to run on the composite's serialized action queue.
+        /// </summary>
+        /// <remarks>
+        /// The action runs at queue-processing time, after any actions enqueued earlier on the
+        /// same queue. Observers can use this to defer advancement decisions until after the
+        /// current synchronous propagation chain has completed -- for example, an applier that
+        /// wants to check a latch that is set by a sibling applier in the same propagation.
+        /// </remarks>
+        /// <param name="action">the action to enqueue</param>
+        void EnqueueAction(Action action);
```
tanderson-ld: Could we instead add a small bit of tagging metadata (FDv2, FDv1) to the entries and then add `void BlockAll(Predicate<(SourceFactory Factory, ActionApplierFactory ActionApplierFactory)> toBlockPredicate).

Then internally, the source list can just block the initializers and synchronizers by the predicate matching the that they are FDv2.  Then the only one that remains are FDv1 whatevers (which is the fallback synchronizer).

This way none of the code related to how it moves through the composite source has to change, the internal queue is not exposed for any action to be enqueued, and there isn't special handling for if FDv1 fallback was seen during the initializers phase or the synchronizers phase.

pkgs/sdk/server/src/Internal/FDv2DataSources/FDv2DataSource.cs
```diff
@@ -364,60 +402,191 @@ public void Apply(ChangeSet<ItemDescriptor> changeSet)
         /// Action applier that blacklists the current datasource when init occurs or when Off status is seen,
         /// then disposes the current datasource, goes to the next datasource, and starts it.
         /// </summary>
-        private class ActionApplierBlacklistWhenSuccessOrOff : IDataSourceObserver
```
tanderson-ld, line 367: This ActionApplierBlacklistWhenSuccessOrOff got extremely complicated.  Is it just easier to make a ActionApplierForInitializers and ActionApplierForSynchronizers and there is a bit of duplicated code wrt handling the FDv1 fallback signal?

I think this approach gets rid of the need for the GatedObserver, the latch, and the deferred re-enqueuing  code.

tanderson-ld (changes requested):

pkgs/sdk/server/src/Internal/FDv2DataSources/FDv2DataSource.cs
```diff
@@ -11,6 +11,7 @@
 namespace LaunchDarkly.Sdk.Server.Internal.FDv2DataSources
 {
     using FactoryList = List<(SourceFactory Factory, ActionApplierFactory ActionApplierFactory)>;
+    using OuterFactoryList = List<(SourceFactory Factory, ActionApplierFactory ActionApplierFactory, CompositeEntryKind Kind)>;
```
tanderson-ld: We don't need to support the list with no kind, we can change the API as this is internal only.

pkgs/sdk/server/src/Internal/DataSources/CompositeDataSource/CompositeSource.cs
```diff
@@ -53,6 +54,21 @@ public CompositeSource(
             IList<(SourceFactory Factory, ActionApplierFactory ActionApplierFactory)> factoryTuples,
             Logger logger,
             bool circular = true)
+            : this(compositeDescription, updatesSink, WithDefaultKind(factoryTuples), logger, circular)
```
tanderson-ld: We don't need to support the constructor with a list with no source kind, we can change the API as this is internal only.

pkgs/sdk/server/src/Internal/DataSources/CompositeDataSource/CompositeSource.cs
```diff
@@ -72,12 +88,24 @@ public CompositeSource(
...
 
-            _sourcesList = new SourcesList<(SourceFactory SourceFactory, ActionApplierFactory ActionApplierFactory)>(
+            _sourcesList = new SourcesList<(SourceFactory Factory, ActionApplierFactory ActionApplierFactory, CompositeEntryKind Kind)>(
                 circular: circular,
                 initialList: factoryTuples
             );
         }
 
+        private static IList<(SourceFactory Factory, ActionApplierFactory ActionApplierFactory, CompositeEntryKind Kind)>
+            WithDefaultKind(IList<(SourceFactory Factory, ActionApplierFactory ActionApplierFactory)> tuples)
```
tanderson-ld: This won't be necessary if we require a list with kinds be provided at construction time (see other comment on constructor)

pkgs/sdk/server/src/Internal/DataSources/CompositeDataSource/SourcesList.cs
```diff
@@ -112,6 +112,38 @@ public int IndexOf(T element)
             return _list.IndexOf(element);
         }
 
+        /// <summary>
+        /// Removes every element that matches the predicate, adjusting the head position so it
+        /// continues to point at whichever element followed the previous head. Returns the number
+        /// of elements removed.
+        /// </summary>
+        public int RemoveAll(Predicate<T> match)
```
tanderson-ld, line 120: Ensure there is sufficient unit test coverage for the various head position edge cases, such as when the predicate matches every entry and all entries are removed.

pkgs/sdk/server/src/Internal/DataSources/CompositeDataSource/SourcesList.cs
```diff
@@ -112,6 +112,38 @@ public int IndexOf(T element)
...
+        {
+            if (match is null) throw new ArgumentNullException(nameof(match));
```
tanderson-ld: We don't need to throw if there is no matcher.  We can just return as the null matcher does not match anything and so no removal occurs.

pkgs/sdk/server/src/Internal/FDv2DataSources/FDv2DataSource.cs
```diff
@@ -375,34 +400,54 @@ public ActionApplierBlacklistWhenSuccessOrOff(ICompositeSourceActionable actiona
...
-                if (newState == DataSourceState.Off)
-                {
-                    _actionable.BlockCurrent();
-                    _actionable.DisposeCurrent();
-                    _actionable.GoToNext();
-                    _actionable.StartCurrent();
-                }
+                if (newState != DataSourceState.Off) return;
+                if (newError?.FDv1Fallback == true) return;
+
```
tanderson-ld, line 409: Bring back comment, it is still a valid comment:

```
// When Off status is seen, blacklist current, dispose current, go to next, and start current
```

pkgs/sdk/server/src/Internal/FDv2DataSources/FDv2DataSource.cs
```diff
@@ -411,18 +456,23 @@ public FDv1FallbackActionApplier(ICompositeSourceActionable actionable)
...
-                // this FDv1 fallback action applier doesn't care about apply, it only looks for the FDv1Fallback flag in the errors
+                if (!changeSet.FDv1Fallback) return;
+                Trigger();
+            }
+
+            private void Trigger()
+            {
+                if (Interlocked.CompareExchange(ref _triggered, 1, 0) != 0) return;
+
+                _actionable.BlockAll(kind => kind == CompositeEntryKind.FDv2);
```
tanderson-ld, line 478: Should use `_actionable.DisposeCurrent();` to be defensive against a source that does not shut itself down.  Since dispose current is supposed to be idempotent, this should not hurt.

pkgs/sdk/server/src/Internal/FDv2DataSources/FDv2StreamingDataSource.cs
```diff
@@ -218,13 +223,32 @@ private void OnMessage(object sender, MessageReceivedEventArgs e)
...
+                        if (fdv1Fallback)
+                        {
+                            _log.Info("LaunchDarkly streaming response indicates fallback to FDv1");
+                            var fallbackError = new DataSourceStatus.ErrorInfo
+                            {
+                                Kind = DataSourceStatus.ErrorKind.Unknown,
+                                Time = DateTime.Now,
+                                FDv1Fallback = true
+                            };
+                            _initTask.TrySetResult(true);
```
tanderson-ld: `_initTask.TrySetResult(true);` being conditional on `if (fdv1Fallback)` seems incorrect.

I think `MaybeMarkInitialized()` just before `if (fdv1Fallback)` took care of this concern.

pkgs/sdk/server/src/Internal/FDv2DataSources/FDv2DataSource.InitializationTracker.cs
tanderson-ld, line 1: Might make sense for @kinyoklion to review this file as he is more familiar with it than I am.

tanderson-ld (changes requested):

tanderson-ld (approved):

PR #264, opened 2026-05-26
fix: gracefully handle MAUI unavailability in non-MAUI Windows apps

tanderson-ld:
**Requirements**

- [ ] I have added test coverage for new or changed functionality.
Not worth the squeeze.

TODO

PlatformConnectivity, AppInfo, and DeviceInfo all call MAUI APIs that require platform initialization. In WPF/WinForms apps targeting net8.0-windows, MAUI is never initialized, causing a TypeInitializationException during LdClient.Init(). Guard the static constructor with a _mauiAvailable flag and fall back to safe defaults (assume connected, return null for app/device info) when MAUI is not available at runtime.

A multilayer approach with default platform impls and a public API for overrides from another package.  This could be a future improvement.

---

> [!NOTE]
> **Low Risk**
> Defensive fallbacks around optional MAUI metadata and connectivity; behavior matches existing netstandard stubs when MAUI is absent, with no auth or data-path changes.
> 
> **Overview**
> Prevents **`LdClient.Init()`** crashes on **net8.0-windows** desktop apps (WPF/WinForms) where MAUI is referenced but never initialized.
> 
> **`PlatformConnectivity`** now sets **`_mauiAvailable`** in a guarded static constructor; when MAUI fails, it assumes **Internet** connectivity, returns **no connection profiles**, and skips subscribing to MAUI connectivity events (aligned with the existing netstandard stub behavior).
> 
> **`AppInfo`** and **`DeviceInfo`** MAUI helpers wrap MAUI API calls in **try/catch** and return **`null`** for app/OS/device metadata when MAUI is unavailable, instead of throwing during init.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 83e2f9b35b7f5d9311ec614c579ea51f8c727af4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkgs/sdk/client/src/PlatformSpecific/AppInfo.maui.cs
+++ b/pkgs/sdk/client/src/PlatformSpecific/AppInfo.maui.cs
@@ -5,10 +5,20 @@ namespace LaunchDarkly.Sdk.Client.PlatformSpecific
 {
     internal static partial class AppInfo
     {
-        internal static ApplicationInfo? GetAppInfo() => new ApplicationInfo(
-            Microsoft.Maui.ApplicationModel.AppInfo.Current.PackageName,
-            Microsoft.Maui.ApplicationModel.AppInfo.Current.Name,
-            Microsoft.Maui.ApplicationModel.AppInfo.Current.BuildString,
-            Microsoft.Maui.ApplicationModel.AppInfo.Current.VersionString);
+        internal static ApplicationInfo? GetAppInfo()
+        {
+            try
+            {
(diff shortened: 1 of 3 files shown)


jsonbailey (approved):
Would it be helpful to log the exceptions as a debug?

tanderson-ld:
> Would it be helpful to log the exceptions as a debug?

Kinda tricky to pipe a logger into this region, I am inclined to not.

PR #266, opened 2026-05-26
feat: Allow register plugins dynamically 

abelonogov-ld:
**Requirements**

Add possibility to Register plugins dynamically for separation concerns from LD flagging and Observability. 
It is not customer facing API and doesn't require documentation change.

```
    var client = await LdClient.InitAsync(ldConfig, context, TimeSpan.FromSeconds(5));
    client.RegisterPlugin(observabilityPlugin);

```             

---

> [!NOTE]
> **Medium Risk**
> Changes hook execution and evaluation/identify wrapping for all clients (unified Executor path) and introduces concurrent hook list updates; failures are logged but mis-registered plugins could affect observability behavior.
> 
> **Overview**
> Adds **`LdClient.RegisterPlugin`** so plugins (e.g. observability) can be attached after `Init`/`InitAsync`, without listing them in `ConfigurationBuilder.Plugins`.
> 
> The hook pipeline is refactored for **runtime hook registration**: `IHookExecutor` gains **`AddHooks`**, and **`Executor`** keeps an immutable **`Stages`** snapshot (hooks + before/after executors) published under a lock; each evaluation/identify call snapshots `_stages` so before/after stages stay consistent if hooks are added mid-flight. With zero hooks, evaluation/identify skip hook work.
> 
> **`NoopExecutor`** is removed—the client always uses **`Executor`** (including when no hooks are configured at startup). **`EnvironmentMetadata`** is always created on the client (not only when config plugins exist), and is reused for dynamic registration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6e1d7660188d3708413d85fc84aed17ca0bc0d36. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkgs/sdk/client/src/LdClient.cs
```diff
@@ -902,6 +902,55 @@ public bool FlushAndWait(TimeSpan timeout) =>
...
+                    plugin.Metadata.Name ?? "unknown", ex);
+            }
+            if (pluginHooks != null && pluginHooks.Count > 0)
+            {
+                _hookExecutor.AddHooks(pluginHooks);
+            }
+
+            try
+            {
+                plugin.Register(this, _environmentMetadata);
```
tanderson-ld, line 945: I see that calling Register after GetHooks fails basically matches existing behavior in LdClient where it uses GetPluginHooks, but it does seem odd to me that it would still call register in the failure case.  Do you have a stronger rationalization for this than just precedent?

tanderson-ld (approved):

PR #277, opened 2026-06-01
feat!: Split LdAiCompletionConfig from LdAiCompletionConfigDefault

jsonbailey:
BEGIN_COMMIT_OVERRIDE
feat!: Split unified LdAiConfig into LdAiCompletionConfig (SDK output) and LdAiCompletionConfigDefault (user input) — Builder, New(), and Disabled move to the Default type; introduce abstract LdAiConfig and LdAiConfigDefault base types for future agent/judge modes
feat!: Change CompletionConfig to return LdAiCompletionConfig instead of ILdAiConfigTracker — obtain a tracker via config.CreateTracker()
feat!: Default LdAiCompletionConfigDefault.Enabled to true per AISDK spec (was false on the LdAiConfig builder in 0.9.x)
feat!: Remove ILdAiConfigTracker.Config property — read config fields from the LdAiCompletionConfig the caller already holds
feat!: Make LdAiConfigTracker SDK-constructed only — the public constructor is removed; obtain trackers via config.CreateTracker() or ILdAiClient.CreateTracker(resumptionToken, context)
feat!: Convert LdAiCompletionConfig and LdAiCompletionConfigDefault from records to classes — equality is reference-based instead of structural
feat!: Remove LaunchDarkly.Sdk.Server.Ai.DataModel namespace — delete unused AiConfig, Meta, Model, Provider, and Message JSON DTO classes
feat: Tolerant LdValue parsing — missing or wrong-typed fields degrade to typed defaults instead of discarding the whole config
feat: Per-message interpolation fallback — a malformed Mustache template keeps raw content for that message rather than discarding the entire config
feat: Mode-mismatch detection — log a warning and return the caller's default when the flag's _ldMeta.mode does not match the requested mode (per sdk-specs#229)
feat: Non-object variation handling — log an error and return the caller's default when the variation result is not an object
END_COMMIT_OVERRIDE

## Summary

This change splits the unified `LdAiConfig` from 0.9.x into two public sealed types so that the data a user supplies as a fallback is clearly separated from the data the SDK returns:

- **`LdAiCompletionConfig`** is what `LdAiClient.CompletionConfig` returns. It is constructed only by the SDK, exposes the evaluated config data, and provides a `CreateTracker()` method that always returns a non-null tracker. There is no public constructor, no public Builder, and no `Disabled` factory.
- **`LdAiCompletionConfigDefault`** is the user-input type passed as the `defaultValue:` parameter. It carries the prompt/model/provider data but has no tracker. Construction uses the public `Builder`, `New()`, and `Disabled` factory.

Tracker construction is driven by a required factory wired up at evaluation time inside `LdAiClient`. `LdAiCompletionConfig.CreateTracker()` delegates to that factory, so callers no longer have to deal with a nullable tracker -- even on fallback paths (parse failure, interpolation failure).

## Type shape (Option B: parallel hierarchies)

Two abstract bases factor out the shared structure, but the Default and Live types do not inherit from each other:

```
LdAiConfigDefault (abstract)         LdAiConfig (abstract)
  Model, Provider, bool? Enabled       Key, Model, Provider, Enabled,
                                       VariationKey, Version,
                                       CreateTracker()
        |                                       |
LdAiCompletionConfigDefault         LdAiCompletionConfig
  + Messages, Builder, New, Disabled  + Messages, CreateTracker
```

Users should reference the concrete sealed types (`LdAiCompletionConfig`, `LdAiCompletionConfigDefault`). Shared payload types (`Message`, `Role`, `ModelConfig`, `ProviderConfig`) live under static `LdAiConfigTypes` in the `Config` namespace (see #284).

This matches the cross-SDK pattern (parallel hierarchies in python and ruby; live-extends-omit-default in js-core). Future `LdAiAgentConfig` / `LdAiJudgeConfig` slot in symmetrically -- each Default extends `LdAiConfigDefault`, each Live extends `LdAiConfig`.

## Why this lands separately from #249

PR #249 contains concurrency fixes for the AI SDK. It rebased on top of this once this merged so that both reviews stay focused.

## Migration

Before (0.9.x):
```csharp
var defaultConfig = LdAiConfig.New().Enable().AddMessage("Hi").Build();
ILdAiConfigTracker tracker = client.Config("flag-key", ctx, defaultConfig);
LdAiConfig cfg = tracker.Config;
```

After (0.10.0):
```csharp
var defaultConfig = LdAiCompletionConfigDefault.New().Enable().AddMessage("Hi").Build();
LdAiCompletionConfig cfg = client.CompletionConfig("flag-key", ctx, defaultConfig);
var tracker = cfg.CreateTracker();
```

The obsolete `Config(...)` method is preserved with the same `[Obsolete]` attribute; only its parameter and return types are updated to match.

## Test plan

- [x] `dotnet build` succeeds across `netstandard2.0`, `net462`, `net8.0`
- [x] `dotnet test --framework net8.0` passes (53/53)
- [x] New test: reflection-based assertion that `LdAiCompletionConfig` has no public constructors (locks in the SDK-only-construction invariant)
- [x] New tests: `CreateTracker()` returns a working tracker on both the parse-failure and interpolation-failure fallback paths
- [ ] Reviewer confirms migration shape matches the cross-SDK pattern

---

> [!NOTE]
> **High Risk**
> Large breaking public API surface (types, return shapes, tracker construction) plus new evaluation/fallback semantics that integrators must migrate to; behavior changes on parse/interpolation errors affect production prompts and telemetry wiring.
> 
> **Overview**
> **Breaking API redesign** for the server AI SDK: unified `LdAiConfig` is replaced by **`LdAiCompletionConfig`** (SDK evaluation result) and **`LdAiCompletionConfigDefault`** (caller fallback with `New`/`Builder`/`Disabled`). `CompletionConfig` now returns the config object; usage tracking is **`config.CreateTracker()`** instead of receiving a tracker directly. `ILdAiConfigTracker.Config` is removed.
> 
> Evaluation logic moves into internal **`ConfigFactory`**, which parses flag JSON as **`LdValue`** (the **`DataModel`** JSON DTOs are deleted). **`LdAiClient`** is slimmed down to `JsonVariation` + factory build. Fallback behavior is more tolerant: wrong type or **`_ldMeta.mode`** mismatch uses the caller’s default (with logging); bad Mustache templates fail **per message** instead of dropping the whole config. Shared shapes live on **`LdAiConfigBase`** / **`LdAiConfigDefaultBase`** with top-level **`Message`**, **`ModelConfig`**, **`ProviderConfig`**, **`Role`**. Default builder **`Enabled`** defaults to **true**; main config types are **classes** (reference equality). **`LdAiConfigTracker`** is **internal** and keyed off **`config.Key`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 86ff9c827e8072383eaaabf22c1f3efa40d5c387. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkgs/sdk/server-ai/src/Config/ConfigFactory.cs
```diff
@@ -0,0 +1,263 @@
...
+/// follow the same pattern) can synthesize a tracker factory and merge in
+/// context-derived prompt variables without the public <see cref="LdAiClient"/>
+/// having to know any of those details.
+/// </summary>
+internal sealed class ConfigFactory
+{
+    // This is the special Mustache variable that can be used in prompts to access the current
+    // LaunchDarkly context. For example, {{ ldctx.key }} will return the context key.
+    private const string LdContextVariable = "ldctx";
+
```
mattrmc1, line 23: nit: The string "completion" is used several times. Can const here. (Would set a precedent for other modes to come?)

mattrmc1 (approved):
A super minor nit but looks great. Pulled down and tested locally 👍

pkgs/sdk/server-ai/src/LdAiClient.cs
```diff
@@ -52,73 +47,15 @@ public LdAiClient(ILaunchDarklyClient client)
...
-                        $"AI model config prompt has malformed message at index {i}: {ex.Message} (returning default config, which will not contain interpolated prompt messages)");
-                    return new LdAiConfigTracker(_client, key, defaultValue, context);
-                }
-            }
-        }
+        defaultValue ??= LdAiCompletionConfigDefault.Disabled;
 
-        return new LdAiConfigTracker(_client, key, new LdAiConfig(parsed.Meta?.Enabled ?? false, prompt, parsed.Meta, parsed.Model, parsed.Provider), context);
+        var ldValue = _client.JsonVariation(key, context, defaultValue.ToLdValue());
+        return _factory.BuildCompletionConfig(key, ldValue, context, variables);
```
kinyoklion: It appears that if this is called for a flag which actually isn't an AI config, then it would return a default constructed completion config instead of the default completion config.

kinyoklion: When the flag exists but its value isn't an AI Config object (e.g. a mistyped variation), `JsonVariation` returns that non-object `LdValue` and it flows straight into `BuildCompletionConfig`. `ParseMeta` then reads `_ldMeta` off a non-object as `LdValue.Null`, so `enabled` degrades to `false` and `mode` defaults to `"completion"` — the result is a default-constructed *disabled, empty* config, and the caller's supplied default is silently dropped (no log either).

This is also a regression vs. `main`, where `ParseConfig` deserialized the value, threw on a non-object, logged, and returned the caller's default. The AISDK spec calls for this explicitly — Req 1.2.3.4 step 2: on a non-dictionary variation result, log an error and *replace the result with the provided default value*.

Here's a unit test (real client + `TestData`, flag set to a bare number) that pins the desired behavior. It **fails on the current code** (`Enabled` comes back `false` and `Messages` is empty):

```csharp
[Fact]
public void ReturnsDefaultConfigWhenFlagIsNotAnAiConfig()
{
    // A flag that exists but whose value is not an AI Config object. Here it is a
    // bare number, which is what JsonVariation hands back when a flag's variation is
    // mistyped. The AI SDK must detect the non-object shape and fall back to the
    // caller's default, rather than tolerantly parsing it into a default-constructed
    // (disabled, empty) completion config.
    var testData = TestData.DataSource();
    testData.Update(testData.Flag("foo").ValueForAll(LdValue.Of(42)));

    var client = new LdClientAdapter(new LdClient(
        Configuration.Builder("key")
            .DataSource(testData)
            .Events(Components.NoEvents)
            .Build()));
    var aiClient = new LdAiClient(client);

    var defaultConfig = LdAiCompletionConfigDefault.New().AddMessage("Hello").Build();

    var result = aiClient.CompletionConfig("foo", Context.New("user-key"), defaultConfig);

    // The returned config must be the caller's default, not a default-constructed config.
    Assert.True(result.Enabled);
    Assert.Collection(result.Messages,
        message =>
        {
            Assert.Equal("Hello", message.Content);
            Assert.Equal(Role.User, message.Role);
        });
}
```

Current failure:
```
Failed LaunchDarkly.Sdk.Server.Ai.LdAiClientTest.ReturnsDefaultConfigWhenFlagIsNotAnAiConfig
  Assert.True() Failure
  Expected: True
  Actual:   False
```

Fix direction: in `BuildCompletionConfig`, if `ldValue.Type != LdValueType.Object`, log an error and re-parse over `defaultValue.ToLdValue()` (which means plumbing `defaultValue` into the factory — it isn't passed today).

kinyoklion (approved):

PR #279, opened 2026-06-03
chore: Add Agent and Judge field parsers to ConfigFactory

mattrmc1:
## Summary

Adds the **parsing primitives** for agent and judge config fields to `ConfigFactory`, plus a small visibility fix on `LdAiConfigBase`. The only new public type is `ToolConfig`.

Four new internal static helpers on `ConfigFactory`:

- `ParseInstructions(LdValue)` — returns the string at `instructions`, or null if missing/wrong-typed (used by agent configs).
- `ParseTools(LdValue)` — reads the `tools` object as a `IReadOnlyDictionary<string, ToolConfig>`, keyed by the wire-format property name. Returns an empty dictionary for missing/wrong-typed inputs (tolerant, consistent with the existing parsers).
- `ParseJudgeConfiguration(LdValue)` — reads the `judgeConfiguration.judges[]` array into a `JudgeConfigurationData` of `(key, samplingRate)` entries. Returns null when the block is missing, an empty list when `judges` is missing/wrong-typed.
- `ParseEvaluationMetricKey(LdValue)` — returns the string at `evaluationMetricKey`, or null.

`ToolConfig` is the only new public type. It is a sealed record with an internal constructor and lives in `LaunchDarkly.Sdk.Server.Ai.Config` alongside `Message`, `ModelConfig`, and `ProviderConfig` — same convention as the rest of the data shapes.

`JudgeConfigurationData` and `JudgeEntry` are internal-only; they're shaped to fit how the judge sampling logic will consume them.

## Visibility fix on `LdAiConfigBase`

`VariationKey` and `Version` on `LdAiConfigBase` are now `internal` rather than `public`. The doc comment on both already said *"This field meant for internal LaunchDarkly usage"* — they were inadvertently exposed. This is a breaking change for any caller that was reading them off a concrete config (`LdAiCompletionConfig`, etc.), but no public API was ever expected to depend on them.

## Test plan

- [ ] New `ConfigFactoryParserTest` covers each parser:
  - `ParseTools_ReturnsTwoEntries_WithCorrectCustomParameters`
  - `ParseJudgeConfiguration_ReturnsTwoEntries_WithCorrectKeysAndSamplingRates`
  - `ParseEvaluationMetricKey_ReturnsParsedString`
  - `ParseInstructions_ReturnsRawUninterpolatedString`
  - `ParseAllFields_WithNoNewFields_ReturnsNullOrEmptyWithoutError` (tolerant fallback for missing blocks)
- [ ] Reviewer confirms `ToolConfig` shape matches the cross-SDK pattern

---

> [!NOTE]
> **Medium Risk**
> Public API break on LdAiConfig.VariationKey/Version for external callers; new parsers are isolated but judge/tool shapes will underpin later agent/judge behavior.
> 
> **Overview**
> Adds **internal `ConfigFactory` parsers** for upcoming agent/judge config modes: `instructions`, root-level `tools`, `judgeConfiguration.judges`, and `evaluationMetricKey`, with tolerant skipping of malformed entries and **no wiring into build paths yet** (step 1 of 6).
> 
> Introduces public **`LdAiConfigTypes.Tool`** and **`JudgeConfiguration`** (+ nested `Judge`) to hold parsed wire data; adds **`ConfigFactoryParserTest`** for happy paths, malformed inputs, and missing fields.
> 
> **Breaking visibility change:** `LdAiConfig.VariationKey` and `Version` are now **`internal`** (still used inside the SDK for trackers), aligning with their “internal LaunchDarkly usage” docs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e7858b0f556777c20a816bf6253641d6a8a94a06. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkgs/sdk/server-ai/src/Config/ConfigFactory.cs
```diff
@@ -195,6 +195,52 @@ private static IReadOnlyList<Message> ParseMessages(LdValue messagesValue)
         return result;
     }
 
+    internal static string ParseInstructions(LdValue value)
+    {
+        var instructions = value.Get("instructions");
+        return instructions.Type == LdValueType.String ? instructions.AsString : null;
+    }
+
+    internal static IReadOnlyDictionary<string, ToolConfig> ParseTools(LdValue toolsValue)
```
jsonbailey: toolsValue is consistent with the other methods I created, but not with your other methods. I think the naming was misleading on mine and really all of the methods should probably either be value, or configValue to indicate they are parsing a value from the top level configValue.

pkgs/sdk/server-ai/src/Config/ConfigFactory.cs
```diff
@@ -195,6 +195,52 @@ private static IReadOnlyList<Message> ParseMessages(LdValue messagesValue)
...
+        {
+            var tool = kv.Value;
+            result[kv.Key] = new ToolConfig(
+                tool.Get("name").AsString ?? "",
+                tool.Get("description").AsString,
+                tool.Get("type").AsString,
+                LdValueObjectToDictionary(tool.Get("parameters")),
+                LdValueObjectToDictionary(tool.Get("customParameters")));
+        }
+        return result;
```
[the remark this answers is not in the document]
jsonbailey: This should be addressed

pkgs/sdk/server-ai/src/Config/ConfigFactory.cs
jsonbailey: Ignore me, thats not correct. We do pass in specific values so they are named accurately.

pkgs/sdk/server-ai/src/Config/ConfigFactory.cs
```diff
@@ -195,6 +195,52 @@ private static IReadOnlyList<Message> ParseMessages(LdValue messagesValue)
...
+    }
+
+    internal static JudgeConfigurationData ParseJudgeConfiguration(LdValue value)
```
jsonbailey: Any reason we are not passing in the judgeSpecific config like we do for other parsers?

pkgs/sdk/server-ai/src/Config/JudgeConfigurationData.cs
```diff
@@ -0,0 +1,25 @@
...
+    public double SamplingRate { get; }
+
+    internal JudgeEntry(string key, double samplingRate)
+    {
+        Key = key;
+        SamplingRate = samplingRate;
+    }
+}
+
+internal sealed class JudgeConfigurationData
```
jsonbailey:
```suggestion
internal sealed class JudgeConfiguration
```

pkgs/sdk/server-ai/src/Config/JudgeConfigurationData.cs
```diff
@@ -0,0 +1,25 @@
+using System.Collections.Generic;
+
+namespace LaunchDarkly.Sdk.Server.Ai.Config;
+
+internal sealed class JudgeEntry
```
jsonbailey:
```suggestion
internal sealed class Judge
```

jsonbailey (changes requested):
Just some minor changes. I would also remove the commit override message from the PR since this is a chore.

pkgs/sdk/server-ai/src/Config/ConfigFactory.cs
mattrmc1: This is the property for attached judges that will be supported on both completion and agent configs

jsonbailey (approved):

PR #280, opened 2026-06-03
chore: Add LdAiAgentConfig and LdAiJudgeConfig types with ConfigFactory build methods (2 of 6)

mattrmc1:
## Summary

Adds the **typed config shapes and factory build methods** for agents and judges. With this change, `ConfigFactory` can produce fully-evaluated `LdAiAgentConfig` and `LdAiJudgeConfig` objects from a flag's JSON.

Four new public types slot into the existing `LdAiConfigBase` / `LdAiConfigDefaultBase` parallel hierarchy:

- **`LdAiAgentConfig`** extends `LdAiConfigBase`. Adds `Instructions` (interpolated) and `Tools`. Internal constructor — SDK-built only.
- **`LdAiAgentConfigDefault`** extends `LdAiConfigDefaultBase`. Public `Builder` with `SetInstructions`, `SetModelParam`, `SetCustomModelParam`, `SetModelName`, `SetModelProviderName`, `Enable` / `Disable`. `New()` factory and `Disabled` helper. `ToLdValue()` serializes to the `_ldMeta { enabled, mode }` envelope that the SDK reads, so a caller default round-trips through `JsonVariation`'s fallback path correctly.
- **`LdAiJudgeConfig`** extends `LdAiConfigBase`. Adds `Messages` and `EvaluationMetricKey`.
- **`LdAiJudgeConfigDefault`** mirrors the agent default, but the builder takes `AddMessage(content, role)` instead of `SetInstructions`.

Each config type carries a `Mode` constant (`"agent"` / `"judge"`) used by its build method to validate `_ldMeta.mode` and fall back to the caller's default on mismatch.

### Factory build methods

`ConfigFactory.BuildAgentConfig` and `BuildJudgeConfig` follow the same shape as the existing `BuildCompletionConfig`:

1. Merge user variables with the LD context (the `ldctx` key).
2. Synthesize a tracker factory bound to that context.
3. Reject non-object variations → log error, return caller's default.
4. Reject mode mismatch → log warning, return caller's default.
5. Otherwise parse model/provider plus the type-specific fields (`instructions` + `tools` for agents; `messages` + `evaluationMetricKey` for judges) and interpolate Mustache templates.

A new private `InterpolateInstructions` helper mirrors the per-field tolerance of the existing `InterpolateMessages`: a malformed template in `instructions` keeps the raw string and logs a warning rather than dropping the whole config.

### Defaults

Both new default builders set `_enabled = true` initially. The base type's `Enabled` is `bool?` so the round-trip through `ToLdValue()` → `JsonVariation` fallback can distinguish "unset" from "explicitly disabled".

## Test plan

- [ ] `LdAiAgentConfigTest` covers happy path, mode mismatch, non-object variation, instructions interpolation, and `LdAiAgentConfigDefault.ToLdValue` round-trip
- [ ] `LdAiJudgeConfigTest` covers the same matrix for judges plus default-type behavior
- [ ] Reviewer confirms parallel hierarchy lines up with the cross-SDK pattern (python, ruby) and that builder method names match the Completion side

---

> [!NOTE]
> **Low Risk**
> Additive SDK surface and factory logic parallel to existing completion handling; extensive tests and safe fallbacks on bad variations.
> 
> **Overview**
> Adds **agent** and **judge** AI config types to the server-ai SDK, mirroring the existing completion config pattern.
> 
> **`ConfigFactory`** gains `BuildAgentConfig` and `BuildJudgeConfig`, which parse flag JSON (`_ldMeta.mode` of `agent` / `judge`), merge Mustache variables with LD context, attach trackers, and fall back to caller defaults on non-object variations or mode mismatch. Agents get interpolated **instructions**, **tools**, and optional **judgeConfiguration**; judges get **messages** and **evaluationMetricKey**. **`InterpolateInstructions`** applies the same malformed-template tolerance as message interpolation.
> 
> New public types: **`LdAiAgentConfig`** / **`LdAiAgentConfigDefault`** (builder with `SetInstructions`, `SetJudgeConfiguration`, model fields; `ToLdValue()` for `JsonVariation` fallback) and **`LdAiJudgeConfig`** / **`LdAiJudgeConfigDefault`** (builder with `AddMessage`, metric key). Unit tests cover happy paths, fallbacks, interpolation, judge config round-trip, and default serialization.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit daaa6482ab2b5215b2e4b14f987fa5a7274f6285. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkgs/sdk/server-ai/src/Config/ConfigFactory.cs
```diff
@@ -93,6 +91,152 @@ private LdAiCompletionConfig BuildCompletionFromDefault(
...
+            _logger.Warn(
+                "AI Config mode mismatch for {0}: expected {1}, got {2}. Returning caller's default.",
+                key, LdAiAgentConfig.Mode, mode);
+            return BuildAgentFromDefault(key, defaultValue, mergedVars, trackerFactory);
+        }
+
+        var model = ParseModel(ldValue.Get("model"));
+        var provider = ParseProvider(ldValue.Get("provider"));
+        var tools = ParseTools(ldValue.Get("tools"));
+        var instructions = InterpolateInstructions(ParseInstructions(ldValue), mergedVars, key);
```
jsonbailey: This will likely cascade from pr 1 but we should send ldValue.Get("instructions") here to make it specific for instructions like the other methods.

pkgs/sdk/server-ai/src/Config/ConfigFactory.cs
```diff
@@ -93,6 +91,152 @@ private LdAiCompletionConfig BuildCompletionFromDefault(
...
+        LdAiAgentConfigDefault defaultValue,
+        IReadOnlyDictionary<string, object> mergedVars,
+        Func<LdAiConfigBase, ILdAiConfigTracker> trackerFactory)
+    {
+        var instructions = InterpolateInstructions(defaultValue.Instructions, mergedVars, key);
+        return new LdAiAgentConfig(
+            key,
+            defaultValue.Enabled ?? true,
+            variationKey: "",
+            version: 0,
```
jsonbailey:
```suggestion
            version: 1,
```

jsonbailey (approved):

PR #292, opened 2026-06-16
feat: Add AgentGraph support to the AI SDK

mattrmc1:
## Summary

Adds first-class support for **agent graphs** to `LaunchDarkly.ServerSdk.Ai`. An agent graph is a LaunchDarkly-managed directed graph whose nodes are existing `LdAiAgentConfig`s and whose edges describe handoffs between agents. The SDK fetches the graph flag, resolves every referenced agent config, validates connectivity, and hands the caller a typed object they can traverse and instrument — without exposing the underlying flag shape.

### `AgentGraph` and `CreateGraphTracker` on `ILdAiGraphClient`

```csharp
// New interface — does not modify ILdAiClient
public interface ILdAiGraphClient
{
    AgentGraphDefinition AgentGraph(
        string graphKey,
        Context context,
        IReadOnlyDictionary<string, object> variables = null);

    AiGraphTracker CreateGraphTracker(string resumptionToken, Context context);
}
```

`LdAiClient` now implements both `ILdAiClient` and the new `ILdAiGraphClient`. The existing `ILdAiClient` interface is **unchanged** — no members added, removed, or modified. Graph functionality lives entirely on the separate `ILdAiGraphClient` interface, making this a purely additive change.

`AgentGraph` fires a `$ld:ai:usage:agent-graph` usage event, evaluates the graph flag, fetches every child `LdAiAgentConfig`, runs connectivity validation, and returns an `AgentGraphDefinition`. The returned definition's `Enabled` property reflects the result of **all** validation checks combined: `_ldMeta.enabled`, root present, every node reachable from the root, and every child agent config fetchable and enabled. When any check fails, the SDK logs a warning and returns a disabled definition whose traversal/inspection methods are safe no-ops; only `GetConfig()` (raw flag value + `_ldMeta`) and `CreateTracker()` remain meaningful.

`CreateGraphTracker` reconstructs an `AiGraphTracker` from a base64url resumption token, enabling cross-process continuation (e.g. emitting graph-level events from a worker that didn't run the original evaluation).

### `AgentGraphDefinition`

```csharp
public AgentGraphNode RootNode();
public AgentGraphNode GetNode(string nodeKey);
public IReadOnlyList<AgentGraphNode> GetChildNodes(string nodeKey);
public IReadOnlyList<AgentGraphNode> GetParentNodes(string nodeKey);
public IReadOnlyList<AgentGraphNode> TerminalNodes();
public AgentGraphFlagValue GetConfig();
public AiGraphTracker CreateTracker();

public void Traverse(
    Func<AgentGraphNode, Dictionary<string, object>, object> fn,
    Dictionary<string, object> initialContext = null);

public void ReverseTraverse(
    Func<AgentGraphNode, Dictionary<string, object>, object> fn,
    Dictionary<string, object> initialContext = null);
```

`Traverse` does a breadth-first walk from the root; `ReverseTraverse` walks from terminals back toward the root (the root is always processed last unless it is the only node). Both are cycle-safe — each node is visited at most once. The callback receives the current node and an accumulator dictionary; whatever the callback returns is stored in the accumulator under that node's key and is visible to subsequent visits.

### Node and edge types

- **`AgentGraphNode`** — wraps an `LdAiAgentConfig` plus its outgoing `GraphEdge`s and exposes `IsTerminal` (true when there are no outgoing edges). Per-node metrics are recorded via the wrapped config's existing `CreateTracker()`.
- **`GraphEdge`** — `record` carrying the target `Key` and an optional `Handoff` dictionary (`IReadOnlyDictionary<string, LdValue>`). Handoff maps are wrapped in `ReadOnlyDictionary` to enforce immutability; edge lists are frozen via `AsReadOnly()`.
- **`AgentGraphFlagValue` / `LdMeta`** — the parsed raw flag value, including `_ldMeta`. Always non-null on the returned definition, even when the graph is disabled.

### `AiGraphTracker`

```csharp
public void TrackInvocationSuccess();       // at-most-once (shares slot with TrackInvocationFailure)
public void TrackInvocationFailure();       // at-most-once
public void TrackDuration(double durationMs); // at-most-once
public void TrackTotalTokens(Usage tokens); // at-most-once (skips slot claim when usage is empty)
public void TrackPath(IReadOnlyList<string> path); // at-most-once

public void TrackRedirect(string sourceKey, string redirectedTarget); // multi-fire
public void TrackHandoffSuccess(string sourceKey, string targetKey);  // multi-fire
public void TrackHandoffFailure(string sourceKey, string targetKey);  // multi-fire

public string ResumptionToken { get; }
public AiGraphMetricSummary Summary { get; }
public AiGraphTrackData GetTrackData();
public static AiGraphTracker FromResumptionToken(string token, ILaunchDarklyClient client, Context context);
```

Graph-scoped tracker emitting `$ld:ai:graph:*` events. At-most-once methods use `Interlocked.CompareExchange` so the contract holds under racing callers — a second call logs a warning and is silently dropped. `TrackTotalTokens` short-circuits before claiming the slot when usage is empty, consistent with `LdAiConfigTracker.TrackTokens`. Multi-fire methods may be called once per edge traversal. Every event carries the standard `runId` + `graphKey` + `version` track data (plus `variationKey` when non-empty). `ResumptionToken` is a base64url-encoded JSON payload of those fields; `FromResumptionToken` validates and round-trips it, throwing `ArgumentException` on malformed or missing-required-fields input.

### `graphKey` propagation through `LdAiConfigTracker`

When a per-node tracker is created via `AgentGraph` (rather than via the standalone `AgentConfig` path), the parent graph's key is threaded through to `LdAiConfigTracker` so every per-node event includes `graphKey` in its track data and resumption token. The wire format omits empty optional fields, so existing non-graph trackers continue to round-trip exactly.

### Other touched files

- `ConfigFactory.BuildAgentConfig` and `TrackerFactoryFor` accept an optional `graphKey` parameter (default `null`). The default-path helper `BuildAgentFromDefault` is now `internal` (was `private`) so the graph code path can reuse it. All existing call sites are unchanged.
- `Polyfills/IsExternalInit.cs` enables `init` accessors and positional records on the `net462`/`netstandard2.0` targets used by the new graph types.

## Migration

**None required.** `ILdAiClient` is unchanged — no new members were added. Graph functionality is on the new `ILdAiGraphClient` interface, which `LdAiClient` implements alongside `ILdAiClient`. Existing consumers, including hand-rolled test doubles that implement `ILdAiClient`, will continue to compile and work without modification. The wire format is backward-compatible (the new `graphKey` field on config resumption tokens is omitted when empty).

## Test plan

- [ ] `dotnet build` succeeds across `netstandard2.0`, `net462`, `net6.0`, `net8.0`
- [ ] `dotnet test pkgs/sdk/server-ai/test/LaunchDarkly.ServerSdk.Ai.Tests.csproj --framework net8.0` passes
- [ ] `AgentGraphDefinitionTest` covers node lookup, parent/child/terminal queries, BFS and reverse-BFS traversal, cycle safety, and disabled-graph no-op behavior
- [ ] `AiGraphTrackerTest` covers track-data shape, at-most-once vs. multi-fire semantics, `TrackTotalTokens` empty-usage short-circuit, `ResumptionToken` round-trip, `FromResumptionToken` error handling, and the `Summary` snapshot
- [ ] `LdAiAgentGraphConfigTest` covers `graphKey` propagation into per-node track data and resumption tokens
- [ ] `LdAiClientAgentGraphTest` covers end-to-end `AgentGraph` retrieval and every validation path: disabled `_ldMeta`, missing root, unreachable node, and unfetchable/disabled child agent config — plus the `$ld:ai:usage:agent-graph` usage event
- [ ] Reviewer confirms graph event names, at-most-once semantics, and resumption-token wire format match the cross-SDK contract

---

> [!NOTE]
> **Medium Risk**
> Large additive public API and shared telemetry/resumption-token contracts must stay aligned with other AI SDKs; optional ILogger.Debug is a compile-time break only for custom ILogger implementations outside the package.
> 
> **Overview**
> Adds **agent graph** support to the server AI SDK: `LdAiClient` now also implements **`ILdAiGraphClient`** with `AgentGraph` and `CreateGraphTracker`, while **`ILdAiClient` stays unchanged**.
> 
> `AgentGraph` evaluates a graph flag, parses `root` / `edges` / `_ldMeta`, validates connectivity and that every referenced agent config is enabled, then returns an **`AgentGraphDefinition`** (or a disabled shell with empty nodes when validation fails). The definition exposes node lookup, BFS **`Traverse`** / terminal-up **`ReverseTraverse`**, and **`CreateTracker`** for graph-scoped metrics.
> 
> New types include **`AgentGraphNode`**, **`GraphEdge`** (optional handoff payload), **`AgentGraphFlagValue`**, **`AiGraphTracker`** (`$ld:ai:graph:*` events, at-most-once vs multi-fire handoff/redirect), and **`AiGraphMetricSummary`**. Graph node fetches thread an optional **`graphKey`** into **`LdAiConfigTracker`** track data and resumption tokens when absent from standalone `AgentConfig` calls.
> 
> Supporting tweaks: **`ILogger.Debug`** (+ adapter), optional **`graphKey`** on **`ConfigFactory.BuildAgentConfig`**, and an **`IsExternalInit`** polyfill for older TFMs. Broad xUnit coverage for definition, tracker, client validation paths, and `graphKey` propagation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d64294c263a42c6c270daf3b4197fb3119141fe6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkgs/sdk/server-ai/src/Config/ConfigFactory.cs
+++ b/pkgs/sdk/server-ai/src/Config/ConfigFactory.cs
@@ -102,10 +102,11 @@ internal sealed class ConfigFactory
         LdValue ldValue,
         Context context,
         LdAiAgentConfigDefault defaultValue,
-        IReadOnlyDictionary<string, object> variables)
+        IReadOnlyDictionary<string, object> variables,
+        string graphKey = null)
     {
         var mergedVars = MergeVariables(variables, context);
-        var trackerFactory = TrackerFactoryFor(context);
+        var trackerFactory = TrackerFactoryFor(context, graphKey);
 
         if (ldValue.Type != LdValueType.Object)
         {
@@ -144,7 +145,7 @@ internal sealed class ConfigFactory
             trackerFactory);
     }
 
-    private LdAiAgentConfig BuildAgentFromDefault(
+    internal LdAiAgentConfig BuildAgentFromDefault(
         string key,
         LdAiAgentConfigDefault defaultValue,
         IReadOnlyDictionary<string, object> mergedVars,
@@ -275,7 +276,7 @@ internal sealed class ConfigFactory
         return result;
     }
 
-    private Func<LdAiConfig, ILdAiConfigTracker> TrackerFactoryFor(Context context)
+    private Func<LdAiConfig, ILdAiConfigTracker> TrackerFactoryFor(Context context, string graphKey = null)
     {
         return cfg => new LdAiConfigTracker(
             _client,
@@ -285,7 +286,8 @@ internal sealed class ConfigFactory
             cfg.Version,
             context,
             cfg.Model?.Name,
-            cfg.Provider?.Name);
+            cfg.Provider?.Name,
+            graphKey);
     }
 
     private static (bool Enabled, string VariationKey, int Version, string Mode) ParseMeta(LdValue value)
--- a/pkgs/sdk/server-ai/src/Config/ConfigFactory.cs
+++ b/pkgs/sdk/server-ai/src/Config/ConfigFactory.cs
@@ -296,7 +296,7 @@ internal sealed class ConfigFactory
         var enabled = meta.Get("enabled").AsBool;
         var variationKey = meta.Get("variationKey").AsString ?? "";
         var versionValue = meta.Get("version");
-        var version = versionValue.IsNull ? 1 : versionValue.AsInt;
+        var version = versionValue.IsNull || versionValue.AsInt <= 0 ? 1 : versionValue.AsInt;
         // Default to the completion mode when _ldMeta.mode is missing or non-string: legacy
         // flags predate the mode tag and were always served as completion configs.
         var mode = meta.Get("mode").AsString ?? LdAiCompletionConfig.Mode;
--- a/pkgs/sdk/server-ai/src/Config/ConfigFactory.cs
+++ b/pkgs/sdk/server-ai/src/Config/ConfigFactory.cs
@@ -296,7 +296,7 @@ internal sealed class ConfigFactory
         var enabled = meta.Get("enabled").AsBool;
         var variationKey = meta.Get("variationKey").AsString ?? "";
         var versionValue = meta.Get("version");
-        var version = versionValue.IsNull || versionValue.AsInt <= 0 ? 1 : versionValue.AsInt;
+        var version = versionValue.IsNull ? 1 : versionValue.AsInt;
         // Default to the completion mode when _ldMeta.mode is missing or non-string: legacy
         // flags predate the mode tag and were always served as completion configs.
         var mode = meta.Get("mode").AsString ?? LdAiCompletionConfig.Mode;
--- a/pkgs/sdk/server-ai/src/Graph/AgentGraphDefinition.cs
+++ b/pkgs/sdk/server-ai/src/Graph/AgentGraphDefinition.cs
@@ -160,21 +160,6 @@ public sealed class AgentGraphDefinition
             }
         }
 
-        // If no terminals were seeded (pure cycle — no leaf nodes exist), seed all non-root nodes
-        // so every node is still visited. For a single-node graph (or self-loop), fall back to
-        // seeding root itself as the only node.
-        if (queue.Count == 0 && root != null)
-        {
-            foreach (var n in _nodes.Values)
-            {
-                if (n.Key == root.Key) continue;
-                if (visited.Add(n.Key))
-                    queue.Enqueue(n);
-            }
-            if (queue.Count == 0 && visited.Add(root.Key))
-                queue.Enqueue(root);
-        }
-
         while (queue.Count > 0)
         {
             var node = queue.Dequeue();
@@ -203,7 +188,7 @@ public sealed class AgentGraphDefinition
         }
 
         // Process root last
-        if (root != null && _nodes.Count > 1)
+        if (root != null && _nodes.Count > 1 && visited.Count > 0)
         {
             var result = fn(root, context);
             context[root.Key] = result;
--- a/pkgs/sdk/server-ai/src/Graph/AgentGraphDefinition.cs
+++ b/pkgs/sdk/server-ai/src/Graph/AgentGraphDefinition.cs
@@ -208,7 +208,9 @@ public sealed class AgentGraphDefinition
 
         foreach (var key in allKeys)
         {
-            var config = agentConfigs[key];
+            if (!agentConfigs.TryGetValue(key, out var config))
+                continue;
+
             var outgoingEdges = flagValue.Edges != null && flagValue.Edges.TryGetValue(key, out var edges)
                 ? edges
                 : (IReadOnlyList<GraphEdge>)Array.Empty<GraphEdge>();
--- a/pkgs/sdk/server-ai/src/Graph/AgentGraphDefinition.cs
+++ b/pkgs/sdk/server-ai/src/Graph/AgentGraphDefinition.cs
@@ -1,0 +1,257 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using LaunchDarkly.Sdk.Server.Ai.Config;
+
+namespace LaunchDarkly.Sdk.Server.Ai.Graph;
+
+/// <summary>
+/// Represents a fully-resolved agent graph returned by
+/// <see cref="LdAiClient.AgentGraph"/>. When <see cref="Enabled"/> is false, all
+/// node collections are empty and traversal is a no-op; only <see cref="GetConfig"/>
+/// and <see cref="CreateTracker"/> remain meaningful.
+/// </summary>
+public sealed class AgentGraphDefinition
+{
+    private readonly AgentGraphFlagValue _flagValue;
+    private readonly IReadOnlyDictionary<string, AgentGraphNode> _nodes;
+    private readonly Func<AiGraphTracker> _createTracker;
+
+    /// <summary>
+    /// Whether the graph passed all validation checks. False if the flag's
+    /// <c>_ldMeta.enabled</c> is false, the root is missing, any node is
+    /// unreachable from the root, or any child agent config could not be fetched.
+    /// </summary>
+    public bool Enabled { get; }
+
+    internal AgentGraphDefinition(
+        AgentGraphFlagValue flagValue,
+        IReadOnlyDictionary<string, AgentGraphNode> nodes,
+        bool enabled,
+        Func<AiGraphTracker> createTracker)
+    {
+        _flagValue = flagValue;
+        _nodes = nodes;
+        Enabled = enabled;
+        _createTracker = createTracker;
+    }
+
+    /// <summary>
+    /// Returns the root node of the graph, or null if the graph is disabled or has no root.
+    /// </summary>
+    public AgentGraphNode RootNode() =>
+        string.IsNullOrEmpty(_flagValue?.Root) ? null : GetNode(_flagValue.Root);
+
+    /// <summary>
+    /// Returns the node with the given key, or null if not found.
+    /// </summary>
+    public AgentGraphNode GetNode(string nodeKey)
+    {
+        if (nodeKey == null) return null;
+        return _nodes.TryGetValue(nodeKey, out var node) ? node : null;
+    }
+
+    /// <summary>
+    /// Returns the direct children of the given node by following its outgoing edges.
+    /// Returns an empty list if the node is not found.
+    /// </summary>
+    public IReadOnlyList<AgentGraphNode> GetChildNodes(string nodeKey)
+    {
+        var node = GetNode(nodeKey);
+        if (node == null) return Array.Empty<AgentGraphNode>();
+
+        return node.Edges
+            .Select(edge => GetNode(edge.Key))
+            .Where(n => n != null)
+            .ToList();
+    }
+
+    /// <summary>
+    /// Returns all nodes that have an outgoing edge pointing to the given node key.
+    /// </summary>
+    public IReadOnlyList<AgentGraphNode> GetParentNodes(string nodeKey)
+    {
+        return _nodes.Values
+            .Where(node => node.Edges.Any(edge => edge.Key == nodeKey))
+            .ToList();
+    }
+
+    /// <summary>
+    /// Returns all nodes with no outgoing edges (leaf nodes).
+    /// </summary>
+    public IReadOnlyList<AgentGraphNode> TerminalNodes() =>
+        _nodes.Values.Where(n => n.IsTerminal).ToList();
+
+    /// <summary>
+    /// Returns the raw flag value including LaunchDarkly metadata. Always non-null,
+    /// even when <see cref="Enabled"/> is false.
+    /// </summary>
+    public AgentGraphFlagValue GetConfig() => _flagValue;
+
+    /// <summary>
+    /// Creates a new graph-level tracker for this invocation.
+    /// </summary>
+    public AiGraphTracker CreateTracker() => _createTracker();
+
+    /// <summary>
+    /// Performs a breadth-first traversal of the graph starting from the root node.
+    /// For each visited node, <paramref name="fn"/> is called with the node and the
+    /// accumulated context dictionary. The return value of <paramref name="fn"/> is
+    /// stored in the context under the node's key and passed to subsequent calls.
+    /// Cycle-safe: each node is visited at most once.
+    /// </summary>
+    public void Traverse(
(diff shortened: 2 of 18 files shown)


pkgs/sdk/server-ai/src/Graph/AgentGraphDefinition.cs
```diff
@@ -0,0 +1,266 @@
...
+        {
+            foreach (var n in _nodes.Values)
+            {
+                if (n.Key == root.Key) continue;
+                if (visited.Add(n.Key))
+                    queue.Enqueue(n);
+            }
+            if (queue.Count == 0 && visited.Add(root.Key))
+                queue.Enqueue(root);
+        }
```
jsonbailey: **`ReverseTraverse` diverges from the spec for graphs with no terminal nodes.** This fallback seeds all nodes when no terminals exist, so a pure cycle (`a→b→c→a`) or root self-loop (`a→a`) visits every node. The spec (`AIGRAPH` Req 1.4) says reverse traversal "starts from the graph's **terminal nodes**" — an empty terminal set should be a no-op, which is what both js-core (`AgentGraphDefinition.ts:187-191`) and python (`agent_graph/__init__.py:232-234`) do. Suggest dropping this block and letting an empty terminal set fall through to a no-op (the `ReverseTraverseIsCycleSafe` test would then assert `Empty(visited)`).

A quick proof: with this code `ReverseTraverse` returns `["b","c","a"]` here vs `[]` in both references. Also worth confirming the normal reverse path honors the spec's "start from the **deepest** terminal node" ordering, not just plain reverse-BFS.

pkgs/sdk/server-ai/src/LdAiClient.cs
```diff
@@ -125,4 +130,152 @@ public ILdAiConfigTracker CreateTracker(string resumptionToken, Context context)
...
+        AiGraphTracker TrackerFactory() => new AiGraphTracker(_client, graphKey, version, context, variationKey);
+
+        var disabled = new AgentGraphDefinition(parsed,
+            new Dictionary<string, AgentGraphNode>(),
+            enabled: false,
+            createTracker: TrackerFactory);
+
+        if (parsed.Meta?.Enabled == false)
+        {
+            _client.GetLogger()?.Warn($"agentGraph: graph \"{graphKey}\" is disabled.");
```
jsonbailey: Spec `AIGRAPH` Req 1.5.2 says graph validation-failure logs should be emitted at **`DEBUG`** level, not warn — a disabled graph is a normal configuration state rather than a warning condition, and js-core uses `_logger?.debug(...)` for these. Suggest switching these `Warn(...)` calls (here and at the other validation-failure sites in this method) to debug.

pkgs/sdk/server-ai/src/Config/ConfigFactory.cs
```diff
@@ -294,7 +296,7 @@ private static (bool Enabled, string VariationKey, int Version, string Mode) Par
         var enabled = meta.Get("enabled").AsBool;
         var variationKey = meta.Get("variationKey").AsString ?? "";
         var versionValue = meta.Get("version");
-        var version = versionValue.IsNull ? 1 : versionValue.AsInt;
+        var version = versionValue.IsNull || versionValue.AsInt <= 0 ? 1 : versionValue.AsInt;
```
jsonbailey: Version defaulting should be: if `_ldMeta.version` isn't set, default to `1`; if it is set, use that value as-is with no further coercion. The new `|| versionValue.AsInt <= 0` clause coerces a present `0`/negative to `1`, which both references avoid (js `version ?? 1`, python `.get("version", 1)` — they default only when the field is *absent*), and this applies to all config types (completion/agent/judge), not just graphs. Suggest reverting to `versionValue.IsNull ? 1 : versionValue.AsInt`.

pkgs/sdk/server-ai/src/Graph/AiGraphTracker.cs
```diff
@@ -0,0 +1,343 @@
...
+
+    /// <summary>
+    /// Records the path of node keys visited during graph execution. At-most-once.
+    /// </summary>
+    public void TrackPath(IReadOnlyList<string> path)
+    {
+        if (path == null) return;
+
+        if (Interlocked.CompareExchange(ref _path,
+                new StrongBox<IReadOnlyList<string>>(path), null) != null)
```
jsonbailey: `TrackPath` stores the caller's `IReadOnlyList<string>` reference directly, and `Summary` returns it (`:109`). If a caller passes a mutable `List<string>` and mutates it afterward, the recorded summary changes underneath. js-core defensive-copies (`[...path]`). Suggest snapshotting, e.g. `new StrongBox<IReadOnlyList<string>>(path.ToArray())`.

pkgs/sdk/server-ai/test/AiGraphTrackerTest.cs
```diff
@@ -0,0 +1,427 @@
...
+                v.Get("path").Count == 3 &&
+                v.Get("path").Get(0).AsString == "agent-a" &&
+                v.Get("path").Get(1).AsString == "agent-b" &&
+                v.Get("path").Get(2).AsString == "agent-c"),
+            1.0), Times.Once);
+    }
+
+    // Test 19: At-most-once — second TrackDuration logs warning and drops
+    [Fact]
+    public void TrackDurationAtMostOnce()
```
jsonbailey, line 128: A few at-most-once paths are unasserted alongside this one: `TrackPath` and `TrackTotalTokens` called twice, and the "empty `Usage` does not consume the slot" ordering (`AiGraphTracker.cs:232-236` — an empty call should still let a later real call fire). Worth adding to lock down the subtle slot semantics.

jsonbailey (commented):
Reviewed the graph implementation against the shipped js-core and python-server-sdk-ai SDKs and reconciled the differences against the AI SDK spec (`sdk-specs`, `specs/AISDK-ai-sdk/`). Security review came back clean (resumption-token decode, at-most-once concurrency, fail-closed validation, graphKey back-compat all verified).

Two spec-compliance items to fix (reverse-traverse no-terminal behavior, validation log level) plus an out-of-scope version-coercion change and a couple of low-severity items inline. Where js-core and python disagreed, the spec sided with js-core on forward-traverse ordering (BFS) and `trackPath` at-most-once, so .NET is correct on those — no change needed there.

jsonbailey (changes requested):

pkgs/sdk/server-ai/src/Config/ConfigFactory.cs
mattrmc1: Good point. I think this was a cursor bugbot fix

pkgs/sdk/server-ai/src/LdAiConfigTracker.cs
```diff
@@ -414,7 +425,7 @@ public static LdAiConfigTracker FromResumptionToken(string token, ILaunchDarklyC
         }
 
         return new LdAiConfigTracker(client, payload.RunId, payload.ConfigKey,
-            payload.VariationKey, payload.Version, context, "", "");
+            payload.VariationKey, payload.Version > 0 ? payload.Version.Value : 1, context, "", "", payload.GraphKey);
```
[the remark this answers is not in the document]
mattrmc1: See: https://github.com/launchdarkly/dotnet-core/pull/292#discussion_r3431719220

pkgs/sdk/server-ai/test/AgentGraphDefinitionTest.cs
```diff
@@ -0,0 +1,423 @@
...
+        var context = Context.New("user");
+        var nodes = AgentGraphDefinition.BuildNodes(flagValue, ThreeNodeConfigs());
+        var graph = new AgentGraphDefinition(flagValue, nodes, enabled: true,
+            createTracker: () => new AiGraphTracker(mockClient.Object, "my-graph-key", 1, context));
+
+        var tracker = graph.CreateTracker();
+        Assert.Equal("my-graph-key", tracker.GetTrackData().GraphKey);
+    }
+
+    // Test 35: Cycle-safe — ReverseTraverse on pure cycle visits every node, root last
```
jsonbailey: Nit: this comment is now stale — the test correctly asserts `Assert.Empty(visited)` (no-op for a terminal-less cycle per AIGRAPH 1.4), so the "visits every node, root last" wording describes the old behavior. Suggest updating to something like "pure cycle has no terminal nodes — reverse traversal is a no-op."

pkgs/sdk/server-ai/src/LdAiClient.cs
```diff
@@ -125,4 +130,152 @@ public ILdAiConfigTracker CreateTracker(string resumptionToken, Context context)
...
+        LdMeta meta = null;
+        var metaValue = flagValue.Get("_ldMeta");
+        if (metaValue.Type == LdValueType.Object)
+        {
+            var versionVal = metaValue.Get("version");
+            var enabledVal = metaValue.Get("enabled");
+            meta = new LdMeta
+            {
+                VariationKey = metaValue.Get("variationKey").AsString,
+                Version = versionVal.IsNull || versionVal.AsInt <= 0 ? 1 : versionVal.AsInt,
```
jsonbailey: `version` should be opaque: if `_ldMeta.version` is set, use it as-is; if it's absent, default to `1`. No other coercion. Suggest:
```csharp
Version = versionVal.IsNull ? 1 : versionVal.AsInt,
```
(matches the completion/agent/judge parse in `ConfigFactory.cs:299`.)

pkgs/sdk/server-ai/src/LdAiConfigTracker.cs
jsonbailey: If the token carries a `version`, return it as-is (including `0`); default to `1` only when it's absent. `payload.Version` is `int?`, so:
```csharp
payload.Version ?? 1
```

pkgs/sdk/server-ai/src/Graph/AiGraphTracker.cs
```diff
@@ -0,0 +1,344 @@
...
+        }
+
+        if (payload == null || string.IsNullOrEmpty(payload.RunId) || string.IsNullOrEmpty(payload.GraphKey))
+        {
+            throw new ArgumentException(
+                "Graph resumption token is missing required fields (runId, graphKey)",
+                nameof(token));
+        }
+
+        return new AiGraphTracker(ldClient, payload.GraphKey, payload.Version > 0 ? payload.Version.Value : 1, context,
```
jsonbailey: Same here — return the encoded `version` as-is, default to `1` only when absent:
```csharp
payload.Version ?? 1
```

jsonbailey (commented):
Follow-up on `_ldMeta.version` handling: it should be carried opaquely — if set, use the value as-is; if absent, default to 1; no other coercion. A few spots still clamp a non-positive value to 1. Details inline.

jsonbailey (approved):

pkgs/sdk/server-ai/src/LdAiConfigTracker.cs
```diff
@@ -103,14 +110,18 @@ internal LdAiConfigTracker(ILaunchDarklyClient client, string runId, string conf
...
+        // resumption tokens round-trip exactly for configs that never carried them.
         using var stream = new MemoryStream();
         using (var writer = new Utf8JsonWriter(stream))
         {
             writer.WriteStartObject();
             writer.WriteString("runId", _runId);
             writer.WriteString("configKey", _configKey);
+            if (!string.IsNullOrEmpty(_graphKey))
+            {
+                writer.WriteString("graphKey", _graphKey);
```
jsonbailey, line 123: The resumption token serializes fields as `runId, configKey, graphKey?, variationKey?, version`, but spec AITRACK Req 1.1.19.1 mandates the order `runId, configKey, variationKey, version, graphKey` ("serialized with keys in the order listed above"). Java, Python, and js-core all follow the spec order.

Decode is field-name-based so this still self-round-trips, but it violates the spec's exact-order MUST and means a .NET-emitted token won't byte-match the other SDKs' tokens (breaking cross-SDK token portability). Suggest moving the `graphKey` block to after the `version` write so the order is `runId, configKey, variationKey, version, graphKey`.

mattrmc1: This comment appears to be stale. Verified we're writing in the correct order
`runId → configKey → variationKey → version → graphKey`

Reference: https://github.com/launchdarkly/dotnet-core/blob/main/pkgs/sdk/server-ai/src/LdAiConfigTracker.cs#L110

PR #300, opened 2026-06-29
feat!: extend AiMetrics, MetricSummary, and TrackMetricsOf

mattrmc1:
BEGIN_COMMIT_OVERRIDE
feat!: ILdAiConfigTracker.TrackDuration now takes a double instead of a float — custom ILdAiConfigTracker implementations must update their TrackDuration override signature
feat: Extend AiMetrics with optional ToolCalls and DurationMs (duration override)
feat: Extend MetricSummary with accumulated ToolCalls and ResumptionToken
feat: TrackMetricsOf auto-tracks tool calls, honors DurationMs override, and aligns operation-vs-extractor failure semantics with the other SDKs
END_COMMIT_OVERRIDE

## Summary

Extends `AiMetrics`, `MetricSummary`, and `TrackMetricsOf` to bring the .NET AI SDK to parity with the JS, Python, and Java SDKs for metric tracking.

### `AiMetrics` — new optional fields

```csharp
public AiMetrics(bool success, Usage? tokens = null,
    IReadOnlyList<string> toolCalls = null, double? durationMs = null)
```

`ToolCalls` carries tool keys invoked during the operation. `DurationMs` provides an explicit duration override — when set, `TrackMetricsOf` uses it instead of the stopwatch measurement.

### `MetricSummary` — new optional fields

```csharp
public record struct MetricSummary(
    double? DurationMs,
    Feedback? Feedback,
    Usage? Tokens,
    bool? Success,
    double? TimeToFirstTokenMs,
    IReadOnlyList<string> ToolCalls = null,      // NEW
    string ResumptionToken = null                 // NEW
);
```

`ToolCalls` accumulates keys from `TrackToolCall` invocations. `ResumptionToken` surfaces the tracker's resumption token on the summary.

### `TrackMetricsOf` — tool call auto-tracking & duration override

`TrackMetricsOf` now extracts `ToolCalls` from `AiMetrics` and calls `TrackToolCalls` automatically. It also honors `DurationMs` when present, falling back to the stopwatch value when null.

The error-handling semantics now match the other SDKs: if the **operation** throws, duration + error are tracked and the exception propagates. If the **metrics extractor** throws, duration is tracked but error is NOT — the AI operation itself succeeded.

### `TrackToolCall` — accumulates keys on summary

`TrackToolCall` now appends to an internal list (lock-guarded) so that `Summary.ToolCalls` reflects all tool invocations. The existing per-call `$ld:ai:tool_call` event emission is unchanged.

## Breaking change

`ILdAiConfigTracker.TrackDuration` now takes a `double` instead of a `float`:

```csharp
- public void TrackDuration(float durationMs);
+ public void TrackDuration(double durationMs);
```

`ILdAiConfigTracker` is a public interface. Callers that *invoke* `TrackDuration(...)` are unaffected — a `float` widens implicitly to `double`. Anyone who *implements* `ILdAiConfigTracker` (custom or test trackers) must update their override signature. This aligns the tracker duration type with `MetricSummary.DurationMs`/`TimeToFirstTokenMs` (already `double?`) and avoids the precision loss of the previous `(float)sw.Elapsed.TotalMilliseconds` cast.

## Migration

Update any custom `ILdAiConfigTracker` implementation's `TrackDuration` signature:

Before:
```csharp
public void TrackDuration(float durationMs) { /* ... */ }
```

After:
```csharp
public void TrackDuration(double durationMs) { /* ... */ }
```

No changes are required for code that only calls `TrackDuration`, `TrackMetricsOf`, or uses the SDK-provided tracker.

## How to test

```bash
dotnet test pkgs/sdk/server-ai/test/LaunchDarkly.ServerSdk.Ai.Tests.csproj --framework net8.0
```

9 new test cases covering backward compat, new field population, auto tool call tracking, duration override vs. stopwatch fallback, extractor failure semantics, and summary accumulation.

## Known limitations

The only breaking change is the `ILdAiConfigTracker.TrackDuration` signature (`float` → `double`), which affects custom implementations of the interface — see Migration above. All other changes are additive: existing constructors and call sites compile and behave identically.

---

> [!NOTE]
> **Medium Risk**
> Public API break on `ILdAiConfigTracker.TrackDuration` affects custom implementations; `TrackMetricsOf` error/duration semantics changed in ways that alter emitted metrics for extractor failures.
> 
> **Overview**
> Brings the .NET server AI SDK’s metric tracking in line with the JS, Python, and Java SDKs.
> 
> **Breaking:** `ILdAiConfigTracker.TrackDuration` now takes `double` instead of `float` (call sites widen implicitly; custom interface implementations must update). Duration is recorded without casting stopwatch results to `float`.
> 
> **Additive:** `AiMetrics` gains optional `ToolCalls` and `DurationMs`. `MetricSummary` adds accumulated `ToolCalls` (from `TrackToolCall`) and `ResumptionToken`. `TrackMetricsOf` auto-emits tool-call events from `AiMetrics.ToolCalls`, uses `DurationMs` when set, stops the timer before the extractor so slow extractors don’t inflate duration, and treats operation failures (duration + error) separately from extractor failures (duration only, no generation error). `TrackToolCall` appends keys under a lock for summary snapshots.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0a17bde55b33f256f92138bbb481bfe24636ba6c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkgs/sdk/server-ai/src/Tracking/AiMetrics.cs
```diff
@@ -16,14 +18,30 @@ public sealed record AiMetrics
...
+    /// <summary>
+    /// Optional list of tool keys invoked during the operation.
+    /// </summary>
+    public readonly IReadOnlyList<string> ToolCalls;
+
+    /// <summary>
+    /// Optional duration override in milliseconds. When set, this value is used instead
+    /// of the stopwatch measurement in <c>TrackMetricsOf</c>.
+    /// </summary>
+    public readonly double? DurationMs;
```
tanderson-ld, line 30: Is this supposed to be a double? I see float uses elsewhere for tracking duration within this PR.

At the same time, it is unusual to see a fractional types use for milliseconds.  Shouldn't we be using the SI unit of second for durations in our system?  Perhaps this decision is long past.

tanderson-ld (approved):

tanderson-ld:
Jason should be able to review before you merge.

pkgs/sdk/server-ai/src/Tracking/AiMetrics.cs
mattrmc1: Thoughts, @jsonbailey?

On the first point, I think `double` is the right call. `Stopwatch` returns `double` natively, and the internal `_durationMs` is already a `StrongBox<double>`, so the `float` gets widened on storage anyway. Moving `trackDuration` to `double` for consistency would be a breaking change, so I split it out here: https://github.com/launchdarkly/dotnet-core/pull/306. Since we're still pre-release, it may be worth resolving now rather than later.

On the second point (SI units), that ship seems to have sailed — but I could be wrong.

pkgs/sdk/server-ai/src/LdAiConfigTracker.cs
```diff
@@ -134,13 +138,26 @@ private string BuildResumptionToken()
...
-        _timeToFirstTokenMs?.Value
-    );
+    public MetricSummary Summary
+    {
+        get
+        {
+            IReadOnlyList<string> toolCalls;
+            lock (_toolCallKeys)
+            {
+                toolCalls = _toolCallKeys.Count > 0 ? _toolCallKeys.AsReadOnly() : null;
```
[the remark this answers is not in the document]
jsonbailey: this should create a new list that locks in the summary at the time of calling it.

pkgs/sdk/server-ai/src/Tracking/AiMetrics.cs
jsonbailey: After talking with Todd, I think we should keep this as a double, but update the other methods to use double instead of float (which I see you already have a PR open for).

jsonbailey (changes requested):
Take a look at the tool call list comment, other than that this seems okay.

pkgs/sdk/server-ai/src/LdAiConfigTracker.cs
```diff
@@ -134,16 +139,29 @@ private string BuildResumptionToken()
...
-        _timeToFirstTokenMs?.Value
-    );
+    public MetricSummary Summary
+    {
+        get
+        {
+            IReadOnlyList<string> toolCalls;
+            lock (_toolCallKeys)
+            {
+                toolCalls = _toolCallKeys.Count > 0 ? _toolCallKeys.ToList().AsReadOnly() : null;
```
jsonbailey: You will need: 
`using System.Collections.Immutable;`

```suggestion
                toolCalls = _toolCallKeys.Count > 0 ? _toolCallKeys.ToImmutableArray() : null;
```

jsonbailey (approved):

PR #325, opened 2026-07-28
fix!: Make AgentGraph traversal topological

mattrmc1:
## Summary

Makes `AgentGraphDefinition` traversal **topological** and gives each visitor a **dependency-scoped
execution context**, aligning the traversal behavior across the LaunchDarkly AI SDKs.

Previously `Traverse` / `ReverseTraverse` were plain BFS over a single shared, accumulating context
dictionary. That had two problems:

1. **Ordering:** where branches of unequal length converge, the convergence node ran on first
   discovery — before all of its predecessors had run.
2. **Context leakage & mutation:** every callback saw the results of *all* previously-visited nodes
   (including unrelated parallel branches), and results were written back into the caller's
   `initialContext`.

## What changed

`pkgs/sdk/server-ai` — `Graph/AgentGraphDefinition.cs` only (no public API/signature changes):

- **`Traverse`** now visits a node only after **all of its reachable predecessors** have been visited
  (Kahn over in-degree; the root is always released first). On cycles, the unvisited node with the
  lowest remaining in-degree is chosen next, ties broken by discovery order.
- **`ReverseTraverse`** now visits a node only after **all of its reachable descendants** have been
  visited, so the root is visited **last** (Kahn over out-degree, root excluded from cycle-break
  selection). Pure cycles now visit every node instead of being a no-op.
- **Scoped context:** each callback receives a fresh dictionary = `initialContext` plus only that
  node's true dependency results (transitive ancestors forward / descendants reverse). Results are
  kept in a private map keyed by node; unrelated branches and self-loops are excluded, and the
  caller's dictionary is never mutated. Cross-node data flows only through callback return values.
- **Determinism:** discovery order is the graph's BFS encounter order (root first, then declared edge
  order), used only for tie-breaks. Ready-check uses `== 0` to match the reference algorithm and the
  other SDKs.

```csharp
void Traverse(Func<AgentGraphNode, Dictionary<string, object>, object> fn,
              Dictionary<string, object> initialContext = null);
```

## Behavior change (breaking)

Titled `fix!` because visit order and callback-context contents change for graphs with convergent
paths, cycles, or parallel branches. Simple linear graphs are unaffected. Cross-SDK parity with
JS/Python/.NET/Java.

## Tests

- A data-driven test over the canonical cross-SDK vectors `G1`–`G6` (+`G2b`) asserts **exact** visit
  order **and exact** context keys in both directions.
- Convergence runs the shared node last; pure cycle visits all nodes (root last); caller context not
  mutated; unrelated branches excluded; self-loop excluded from its own context; deterministic across
  runs.

## Notes

- No manual `CHANGELOG.md` edit — release-please generates it from the conventional-commit title.
- Graph tracking / wire parsing unchanged; this PR is scoped to traversal + context.

---

> [!NOTE]
> **Overview**
> **`Traverse` and `ReverseTraverse`** on `AgentGraphDefinition` no longer use plain BFS over one shared, mutating context dictionary. They now visit reachable nodes in **topological order** (predecessors first forward; descendants first backward, **root last** on reverse), with **cycle-safe** tie-breaking from BFS discovery order.
> 
> Each callback gets a **scoped** dictionary: `initialContext` plus only that node’s true dependency results (transitive ancestors or descendants). Node outputs live in a private map; unrelated parallel branches and self-loops are excluded, and the caller’s `initialContext` is not mutated.
> 
> **Breaking:** visit order and context keys change for convergent paths, cycles, and parallel branches; linear chains stay the same. Aligns traversal with other LaunchDarkly AI SDKs.
> 
> Tests add canonical **G1–G6 / G2b** vectors asserting exact order and context keys in both directions, plus updated cycle and scoped-context expectations.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 12826ecb9cfef24410c13fb04287d6fe58e75f5d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkgs/sdk/server-ai/src/Graph/AgentGraphDefinition.cs
+++ b/pkgs/sdk/server-ai/src/Graph/AgentGraphDefinition.cs
@@ -94,12 +94,13 @@ public sealed class AgentGraphDefinition
     public AiGraphTracker CreateTracker() => _createTracker();
 
     /// <summary>
-    /// Performs a breadth-first traversal of the graph starting from the root node.
-    /// For each visited node, <paramref name="fn"/> is called with the node and the
-    /// accumulated context dictionary. The return value of <paramref name="fn"/> is
-    /// stored in the context under the node's key and passed to subsequent calls.
-    /// Cycle-safe: each node is visited at most once.
+    /// Visits each reachable node in topological order (predecessors first, root first).
+    /// Ties break by discovery order. Cycle-safe.
     /// </summary>
+    /// <remarks>
+    /// <paramref name="fn"/> receives <paramref name="initialContext"/> plus results
+    /// from that node's reachable predecessors only.
+    /// </remarks>
     public void Traverse(
         Func<AgentGraphNode, Dictionary<string, object>, object> fn,
         Dictionary<string, object> initialContext = null)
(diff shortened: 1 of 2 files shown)


pkgs/sdk/server-ai/src/Graph/AgentGraphDefinition.cs
tanderson-ld, line 1: There are a lot of new dictionary calls, toDictionary, as well as several contains and order by.  Do you have any performance requirements wrt throughput or memory allocation delays?  May be worth a quick benchmark to see if your expected cases are well behaved.

I'm guessing this would apply to the other PRs for the other SDKs if they are following similar implementations.

pkgs/sdk/server-ai/src/Graph/AgentGraphDefinition.cs
```diff
@@ -108,91 +109,143 @@ public void Traverse(
...
-        if (_nodes.Count == 0) return;
+        var root = RootNode();
+        if (root == null) return;
 
+        var rootKey = root.Key;
         var context = initialContext ?? new Dictionary<string, object>();
+        var (reachable, order) = ReachableAndDiscovery(rootKey);
+
+        var outdeg = reachable.ToDictionary(
+            k => k, k => GetNode(k).Edges.Count(e => reachable.Contains(e.Key)));
```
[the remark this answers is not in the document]
mattrmc1: Real, but intentional for now: this is a port of the shared cross-SDK algorithm and js-core computes out-degree identically, so fixing it only in .NET would break the parity this PR is for. Edges into the root are cycles, and cycle ordering is deterministic-but-arbitrary by design. Raising the asymmetry with indeg[root] = 0 against the shared spec, plus a test vector with an edge back to the root.

pkgs/sdk/server-ai/src/Graph/AgentGraphDefinition.cs
mattrmc1: No hard requirements here — traversal runs once per graph invocation and every callback makes an LLM call, so we're spending microseconds against seconds. Benchmarked to confirm: a 20-node graph is ~65 µs / 60 KB and a 50-node fan-out ~150 µs, with degradation only past a few hundred nodes, well beyond expected graph sizes.

mattrmc1: Took a look at the space concerns too. The old approach was actually worse since it carried context level to level. Now, if you have a fan out graph, each node gets `initialContext` plus ONLY its transitive ancestors. The old way gave each node all context from levels above as well as sibling context depending on order of BFS traversal. Assuming the the graph is a diamond shape, the bloated context on the exit node is the same either way.

```
        a
       / \
      b   c
      |   |
      d   e
       \ /
        f
```
| node | old context | new context |
| --- | --- | --- |
| a | {} | {} |
| b | {a} | {a} |
| c | {a, b} | {a} |
| d | {a, b, c} | {a, b} |
| e | {a, b, c, d} | {a, c} |
| f | {a, b, c, d, e} | {a, b, c, d, e} |

PR #213, opened 2026-01-06
chore: Additional solution organization and sandbox.

kinyoklion:
This set of changes shouldn't have any production facing impact. It just makes testing, validation, and development easier.

Extend the solution with the following capabilities:

- A DebugLocalReferences build configuration which locally links dependencies when possible. This is currently only redis and dynamo linking to the server SDK. It cannot yet link common locally because of its use by internal. Once internal is added to the solution it can be extended.
- Added documentation items to the individual project that use them as links. This allows more easy usage of things like README, CONTRIBUTING, and API docs from within a solution.
- Added top-level documentation to the repo-docs solution folder.

image

image

image

Added a new sandbox project that is configured with environment variables. So you can use different data system configurations based on environment variables. This includes persistent stores when using the DebugLocalReferences build.

---

> [!NOTE]
> Introduces local-dev build wiring and a new sandbox for easier testing while organizing solution items.
> 
> - Adds `DebugLocalReferences` configuration across projects to switch `LaunchDarkly.ServerSdk` from NuGet `PackageReference` to local `ProjectReference`; defines `DEBUG_LOCAL_REFERENCES` where needed
> - Updates multiple `.csproj` files to include repo docs (`README`, `CHANGELOG`, etc.) as linked content; adds doc source links in server/server-ai/telemetry
> - Adds `sandbox/dotnet-server-sandbox` console app configurable via environment variables for data system modes (`default`, `streaming`, `polling`, `daemon`, `persistent-store`) and persistent stores (Redis/DynamoDB) when using `DebugLocalReferences`
> - Updates solution (`DotnetCore.sln`) to include `sandbox` and `repo-docs` folders and the new configuration entries; adjusts `InternalsVisibleTo` to `dotnet-server-sandbox`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8e2a733a28b3f7f2148f3f2f0f83eedb965572e5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #327, opened 2026-08-05
chore: Repair DebugLocalReferences and link common/internal locally

jsonbailey:
SDK-2866

#213 added a `DebugLocalReferences` build configuration so in-repo packages link to each other by `ProjectReference` instead of NuGet. It could only be applied partially at the time — quoting that PR:

> It cannot yet link common locally because of its use by internal. Once internal is added to the solution it can be extended.

Internal is now in the monorepo, so the remaining work can be finished. Reviewing #213 also turned up two defects that are broken on `main` today.

Four files. No production-facing change: package metadata, dependency versions, and the Debug/Release dependency graphs are untouched.

## 1. `DebugLocalReferences` has never built, since #213 landed

```
dotnet build DotnetCore.sln -c DebugLocalReferences   →  120 errors
```

Confirmed by checking out #213's own merge commit and building it there — same 120 errors, same project, same error codes. It shipped untested; nothing later broke it.

All errors are in `LaunchDarkly.ServerSdk.SharedTests`:

```
CS0433: The type 'IBigSegmentStore' exists in both
  'LaunchDarkly.ServerSdk, Version=7.0.0.0' and
  'LaunchDarkly.ServerSdk, Version=8.16.0.0'
```

The cause is in `DotnetCore.sln`, not the project files. 28 of the `DebugLocalReferences` entries mapped to `Debug|Any CPU` — including `LaunchDarkly.ServerSdk.SharedTests.Tests`, which project-references `SharedTests`. `SharedTests` was therefore restored under *both* configurations into a single `obj/project.assets.json`. The `Debug` pass won and left the NuGet `LaunchDarkly.ServerSdk 7.0.0-alpha.3` in the dependency graph; the compile then ran as `DebugLocalReferences` and added the local project reference on top, putting both assemblies in scope.

Every project in the dependency closure now maps to `DebugLocalReferences`.

## 2. Two test projects were never built by the solution

`LaunchDarkly.ServerSdk.Ai.Tests` and `LaunchDarkly.ServerSdk.Telemetry.Tests` were attached to the solution as `ProjectSection(SolutionItems)` entries under the `server-ai` and `telemetry` folders instead of as `Project(...)` entries, so `dotnet build DotnetCore.sln` skipped them entirely.

This dates to #201, the commit that created the solution — every other test project was added as a real project there; these two are the only exceptions, and they're also the only two solution folders carrying a `SolutionItems` block. Both are now real projects. They pass (241 and 21 tests) — they hadn't rotted, just been invisible to the solution. The per-package CI workflows have been covering them all along, which is why nobody noticed.

## 3. Internal now participates in `DebugLocalReferences`

`LaunchDarkly.InternalSdk.csproj` gains the `<Configurations>` property, the `DEBUG;TRACE` define, a conditional `ProjectReference` to common, and the linked repo docs (README / CHANGELOG / CONTRIBUTING / docfx.json) its siblings already had.

The `DEBUG` define is load-bearing: `pkgs/shared/internal/src/AssemblyInfo.cs` wraps its `InternalsVisibleTo` in `#if DEBUG`, which is why every other project in #213 received that same `DefineConstants` override.

## 4. Server and client SDKs link common and internal locally

Both used NuGet packages for `LaunchDarkly.CommonSdk` and `LaunchDarkly.InternalSdk` in every configuration, so a change to either shared package couldn't be validated against the SDKs without publishing. They now use a `ProjectReference` under `DebugLocalReferences` and keep the `PackageReference` for Debug and Release.

## Verification

Run in freshly-created worktrees at this commit, one per configuration, so no prior build artifacts could affect the result.

| Check | Result |
|---|---|
| `dotnet build DotnetCore.sln -c DebugLocalReferences` | 0 errors (was 120) |
| `dotnet build DotnetCore.sln -c Debug` | 0 errors |
| ServerSdk / ClientSdk / InternalSdk refs under `DebugLocalReferences` | `CommonSdk`, `InternalSdk` → `project` |
| Same under `Debug` | → `package`, versions unchanged |
| `LaunchDarkly.ServerSdk.Tests` | 1595 passed |
| `LaunchDarkly.CommonSdk.Tests` | 273 passed |
| `LaunchDarkly.ServerSdk.Ai.Tests` | 241 passed |
| `LaunchDarkly.InternalSdk.Tests` | 180 passed |
| `LaunchDarkly.ServerSdk.Telemetry.Tests` | 21 passed |

`Release` fails locally on `CS7027` missing `.snk` files, which is pre-existing on `main` — the signing keys are CI secrets.

One note for reviewers: because NuGet's assets file lives at `obj/project.assets.json` and is not per-configuration, switching between `Debug` and `DebugLocalReferences` requires a restore. `dotnet test --no-restore` right after building the other configuration will produce duplicate-type errors. IDEs restore on configuration switch, so this only bites scripted invocations.

## Follow-up

Not in this PR: the client SDK pins `LaunchDarkly.InternalSdk 3.7.0` while the repo builds 3.10.0. It compiles fine against 3.10.0 under the local configuration, but bumping the pin is a behavioral change and gets its own PR.

---

> [!NOTE]
> **Overview**
> Makes `DebugLocalReferences` actually build and lets in-repo packages consume local `CommonSdk` / `InternalSdk` instead of NuGet.
> 
> **Solution mapping.** Many `DebugLocalReferences` entries in `DotnetCore.sln` were mapped to `Debug`, which mixed NuGet and project refs (duplicate types in SharedTests). Every project now maps to `DebugLocalReferences`. AI and Telemetry test projects are real solution projects instead of SolutionItems, so they participate in `dotnet build`.
> 
> **Local refs.** Client, server, and InternalSdk use `ProjectReference` to common/internal only under `DebugLocalReferences`; Debug/Release keep the existing PackageReferences. InternalSdk also gets the `DEBUG;TRACE` define and linked docs so `InternalsVisibleTo` works in that configuration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 171bfdc9246e1c19f02b588f9e50ab99cf855b3d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/DotnetCore.sln
+++ b/DotnetCore.sln
@@ -59,16 +59,10 @@ EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "LaunchDarkly.ServerSdk.SharedTests.Tests", "pkgs\shared\dotnet-server-sdk-shared-tests\test\LaunchDarkly.ServerSdk.SharedTests.Tests.csproj", "{8C9BDA1A-6AB3-4B52-8CBB-6B702FABC292}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "server-ai", "server-ai", "{EF541D97-2C23-4FB8-BD87-ADFDF0FF83F3}"
-	ProjectSection(SolutionItems) = preProject
-		pkgs\sdk\server-ai\test\LaunchDarkly.ServerSdk.Ai.Tests.csproj = pkgs\sdk\server-ai\test\LaunchDarkly.ServerSdk.Ai.Tests.csproj
-	EndProjectSection
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "LaunchDarkly.ServerSdk.Ai", "pkgs\sdk\server-ai\src\LaunchDarkly.ServerSdk.Ai.csproj", "{561F6F3D-09C5-470A-83AC-F0E97BAD609B}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "telemetry", "telemetry", "{FF3153CA-F727-4666-BA1E-5459BE34FBAB}"
-	ProjectSection(SolutionItems) = preProject
-		pkgs\telemetry\test\LaunchDarkly.ServerSdk.Telemetry.Tests.csproj = pkgs\telemetry\test\LaunchDarkly.ServerSdk.Telemetry.Tests.csproj
-	EndProjectSection
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "LaunchDarkly.ServerSdk.Telemetry", "pkgs\telemetry\src\LaunchDarkly.ServerSdk.Telemetry.csproj", "{DE321396-6459-4C02-AF61-1A2677A7CB65}"
 EndProject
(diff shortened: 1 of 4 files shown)


pkgs/shared/internal/src/LaunchDarkly.InternalSdk.csproj
```diff
@@ -25,17 +25,51 @@
...
+    </None>
+    <None Include="..\docfx.json">
+      <Link>docfx.json</Link>
+      <Pack>false</Pack>
+    </None>
+    <None Include="..\README.md">
+      <Link>README.md</Link>
+      <Pack>false</Pack>
+    </None>
+  </ItemGroup>
```
tanderson-ld, line 71: Is this in scope / necessary for the objective of the PR?

DotnetCore.sln
tanderson-ld, line 1: Were these changes generated by tooling or by AI?

tanderson-ld (approved):
Approved assuming near zero risk and DotnetCore.sln changes are from tooling and not just AI.  Perhaps even if they are via AI and confirmed locally on your machine to be proper, then that is fine as well.

pkgs/shared/internal/src/LaunchDarkly.InternalSdk.csproj
jsonbailey: I can pull it into its own if you think its worth it. It was a small change and was more about getting the projects setup properly.

DotnetCore.sln
jsonbailey: Generated by AI. If you do it manually you get a lot of extra unnecessary folder structures. It has been tested and seems to work. It shouldn't impact deployed project and is really a nicety for local development.

jsonbailey: I noticed looking at an old random PR that we held off on these changes because internal was not in the repo which it now is, so I decided to finish what was started.

PR #334, opened 2026-08-19
fix: honor BackgroundPollInterval when app is in background

tanderson-ld:
`PollingDataSourceBuilder.Build()` was always passing `_pollInterval` to the constructed `PollingDataSource`, regardless of `clientContext.InBackground`. Users configuring polling with a distinct `BackgroundPollInterval` had it silently ignored; streaming-configured SDKs that fell back to polling in background used the 5-minute foreground default rather than the 60-minute background default.

The polling builder now selects the correct interval based on `InBackground`. The streaming builder's background fallback also sets both fields to the background value for defense-in-depth.

Fixes [SDK-2920](https://launchdarkly.atlassian.net/browse/SDK-2920).

---

> [!NOTE]
> **Overview**
> Fixes **background polling** so `BackgroundPollInterval` is actually used when `InBackground` is true, instead of always applying the foreground `PollInterval`.
> 
> `PollingDataSourceBuilder.Build()` now picks `_pollInterval` vs `_backgroundPollInterval` based on `clientContext.InBackground`. When streaming falls back to polling in the background, the streaming builder also passes the background interval into `PollInterval` on the nested polling builder (defense in depth).
> 
> An internal `PollingInterval` getter on `PollingDataSource` supports new tests for foreground vs background builds and streaming’s background fallback.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb9e3d46488554d9d1194de1e401563feceef2d4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

[SDK-2920]: https://launchdarkly.atlassian.net/browse/SDK-2920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkgs/sdk/client/test/LaunchDarkly.ClientSdk.Tests/Integrations/PollingDataSourceBuilderTest.cs
+++ b/pkgs/sdk/client/test/LaunchDarkly.ClientSdk.Tests/Integrations/PollingDataSourceBuilderTest.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Reflection;
 using LaunchDarkly.Sdk.Client.Internal.DataSources;
 using LaunchDarkly.TestHelpers;
 using Xunit;
@@ -10,11 +9,6 @@ namespace LaunchDarkly.Sdk.Client.Integrations
     {
         private readonly BuilderBehavior.InternalStateTester<PollingDataSourceBuilder> _tester =
             BuilderBehavior.For(Components.PollingDataSource);
-
-        private static TimeSpan GetPollingInterval(PollingDataSource dataSource) =>
-            (TimeSpan)typeof(PollingDataSource)
-                .GetField("_pollingInterval", BindingFlags.NonPublic | BindingFlags.Instance)
-                .GetValue(dataSource);
 
         [Fact]
         public void BackgroundPollInterval()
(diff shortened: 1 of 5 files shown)


pkgs/sdk/client/test/LaunchDarkly.ClientSdk.Tests/Integrations/PollingDataSourceBuilderTest.cs
```diff
@@ -9,6 +11,11 @@ public class PollingDataSourceBuilderTest
         private readonly BuilderBehavior.InternalStateTester<PollingDataSourceBuilder> _tester =
             BuilderBehavior.For(Components.PollingDataSource);
 
+        private static TimeSpan GetPollingInterval(PollingDataSource dataSource) =>
```
kinyoklion: I would prefer an internal method. The test projects have access using "InternalsVisibleTo".

kinyoklion (changes requested):

pkgs/sdk/client/test/LaunchDarkly.ClientSdk.Tests/Integrations/PollingDataSourceBuilderTest.cs
tanderson-ld: Good call — replaced the reflection helper with an `internal TimeSpan PollingInterval => _pollingInterval;` accessor on `PollingDataSource`, tests now use `dataSource.PollingInterval` directly. Pushed as 707f53d2.

kinyoklion (approved):

